PR #19, opened 2021-02-10
add round up shuffle test

brainwashers:


vkryvytskyy (approved):
@roman-mate we discussed with @brainwashers to change this test. Cause if we check only `left` and `down` buttons, we face the issue when there is possibilities to move only up/right, but tests think that there are no new moves available.

brainwashers:
> @roman-mate we discussed with @brainwashers to change this test. Cause if we check only `left` and `down` buttons, we face the issue when there is possibilities to move only up/right, but tests think that there are no new moves available.

turned out that new boxes don’t appear on the table if indexes are not changed after clicking the arrows.

roman-mate (approved):

PR #24, opened 2021-03-23
Solution

AndriyPT82:
https://andriypt82.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -80,7 +80,7 @@ document.addEventListener('keydown', e => {
     messageLose.classList.remove('hidden');
   }
 
-  if (checkForValue2048(tableRows)) {
+  if (checkForValue2048(matrix)) {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,240 @@
...
+  };
+
+  return true;
+};
+
+function checkForValue2048(table) {
+  return [...table].some(row => {
+    return [...row.children].some(cell => {
+      return cell.innerText === '2048';
+    });
```
vkryvytskyy, line 238: you don't need inner some, use includes

vkryvytskyy: row.includes(2048)

vkryvytskyy (approved):
GJ

src/scripts/main.js
AndriyPT82: Done!

PR #88, opened 2022-08-03
Solution

vyurkov:
[DEMO LINK](https://vyurkov.github.io/js_2048_game/)


Bohdan-Maksymiuk (approved):
It is better to create classes for games and there will be static fields and methods
Good job overall!!! Keep rocking 🥇

PR #107, opened 2022-09-18
Solution

DariaVeretyak:
https://dariaveretyak.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,347 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const tbody = document.querySelector('tbody');
+const tRows = tbody.rows;
+const controls = document.querySelector('.controls');
+const maxCellValue = 2048;
+
+const field = [
+  ['', '', '', ''],
+  ['', '', '', ''],
+  ['', '', '', ''],
+  ['', '', '', ''],
+];
+
+const classes = [2, 4, 6, 8, 16, 32, 64, 128, 256, 512, 1024, 2048];
+
+controls.insertAdjacentHTML('beforeend', `
+  <button class="button restart hidden">Restart</button>
+`);
+
+const restart = controls.querySelector('.restart');
+const start = controls.querySelector('.start');
+
+const gameScore = document.querySelector('.game-score');
+
+const messageStart = document.querySelector('.message-start');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+
+let counter;
+
+// переносим все данные с массива field
+const getValueField = function() {
+  for (let i = 0; i <= 3; i++) {
+    for (let j = 0; j <= 3; j++) {
+      tRows[i].cells[j].innerText = field[i][j];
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,347 @@
...
+  } else {
+    newValue = 2;
+  }
+}
+
+let row;
+let column;
+// проверяем на пустоту
+const selectEmptyCell = () => {
+  row = getRandomCell(0, field.length);
```
ivanchuk-yurii, line 124: `row = getRandomCell`, if this function is used not only for cells, rename it

src/scripts/main.js
```diff
@@ -1,3 +1,347 @@
...
+      };
+    }
+  }
+
+  return randomValue;
+};
+
+let newValue;
+
+function getRandomValue() {
```
ivanchuk-yurii, line 110: return value from this function, it's hard to track variable

src/scripts/main.js
```diff
@@ -1,3 +1,347 @@
...
+    newValue = 4;
+  } else {
+    newValue = 2;
+  }
+}
+
+let row;
+let column;
+// проверяем на пустоту
+const selectEmptyCell = () => {
```
ivanchuk-yurii, line 123: return value from this function, it's hard to track variables

src/scripts/main.js
```diff
@@ -1,3 +1,347 @@
...
+};
+
+// нажатие start
+start.addEventListener('click', (ev) => {
+  addNewValue();
+  addNewValue();
+  getValueField();
+  getFilledCells();
+
+  if (filledCellsArr.length >= 1) {
```
ivanchuk-yurii, line 188: reason of this `if`?

src/scripts/main.js
```diff
@@ -1,3 +1,347 @@
...
+      if (field[i][j] !== '') {
+        cellField = field[i][j];
+        tRows[i].cells[j].classList.add(`field-cell--${cellField}`);
+      }
+    }
+  }
+};
+
+// обнуляем field
+const getEmptyField = () => {
```
ivanchuk-yurii, line 70: why this is called `get`, it should be `resetField`

ivanchuk-yurii (changes requested):
pay attention to variable naming. try to not use global variables when they are needed only in one function. instead use them only in this place as returned value from another function. when I click any button except arrows, numbers are still adding. condition `The merged cell can’t be merged twice during one move` doesn't work if numbers are next to each other. a few bugs when trying to merge numbers on the beginning. try to test in different scenarios 

https://user-images.githubusercontent.com/53611910/191805866-b3952b98-8dda-49ad-a1f9-071d942dfdb1.mov
https://user-images.githubusercontent.com/53611910/191804317-e781d73e-b31f-4694-855c-0d9e07cc2710.mov
https://user-images.githubusercontent.com/53611910/191802706-edf22d95-3890-42ae-b424-08ecbeeb440a.mov

marina-tilniak (approved):

PR #111, opened 2022-09-24
Develop

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,3 +1,60 @@
...
+  arrowsPlayHandler,
+  createItemInRandomEmptyField,
+} = require('./move');
+
+// vars:
+const gameScore = document.querySelector('.game-score');
+const gameField = document.querySelector('.game-field');
+const allCells = gameField.querySelectorAll('.field-cell');
+const startBtn = document.querySelector('.button');
+const highestScore = document.querySelector('.highest-score');
```
OleksiiBlezniuk, line 18: Usually we have a separated file for the variables, where we export them ad import where we need them, but in this case it's not critical since we use querySelector

src/scripts/main.js
```diff
@@ -1,3 +1,60 @@
...
+
+// main function:
+function gameInit() {
+  highestScore.innerText = localStorage.getItem('score') || 0;
```
OleksiiBlezniuk, line 22: It's okay since you saving only one item in localStorage, but usually we have separated method for that purpose that will return item from it (or null if there is nothing there)

src/scripts/main.js
```diff
@@ -1,3 +1,60 @@
...
+
+  startBtn.addEventListener('click', (e) => {
+    if (startBtn.innerText === 'Start') {
+      startBtn.innerText = 'Restart';
+      startBtn.classList.remove(`start`);
+      startBtn.classList.add(`restart`);
+
+      document.querySelector('.message-play').classList.remove('hidden');
+      document.querySelector('.message-start').classList.add('hidden');
```
OleksiiBlezniuk, line 31: Better to create constants for this classes, 'cause now it looks like magic strings

OleksiiBlezniuk: Also you can save result of querySelector in constant, instead of calling this method every time

src/scripts/main.js
```diff
@@ -1,3 +1,60 @@
...
+
+// vars:
+const gameScore = document.querySelector('.game-score');
+const gameField = document.querySelector('.game-field');
+const allCells = gameField.querySelectorAll('.field-cell');
+const startBtn = document.querySelector('.button');
+const highestScore = document.querySelector('.highest-score');
+
+// main function:
+function gameInit() {
```
OleksiiBlezniuk, line 21: Maybe let's move this function to separated file?

src/scripts/mobile.js
```diff
@@ -0,0 +1,43 @@
...
+  } else if (endX > startX && endX - startX >= 100) {
+    e.key = 'ArrowRight';
+    document.body.dispatchEvent(e);
+  } else if (endY < startY && startY - endY >= 100) {
+    e.key = 'ArrowUp';
+    document.body.dispatchEvent(e);
+  } else if (endY > startY && endY - startY >= 100) {
+    e.key = 'ArrowDown';
+    document.body.dispatchEvent(e);
+  }
```
OleksiiBlezniuk, line 38: We can use return in every case, then we'll be able to remove this `else`s

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+  getNotEmptyFieldsCoords,
+  checkBetween,
+  randomField,
+} = require('./utils');
+
+const gameScore = document.querySelector('.game-score');
+const gameField = document.querySelector('.game-field');
+const allCells = gameField.querySelectorAll('.field-cell');
+const highestScore = document.querySelector('.highest-score');
+const values = [2, 2, 2, 2, 2, 2, 2, 2, 2, 4];
```
OleksiiBlezniuk, line 15: Do we really need this array? 🙂 As far as I understand it was used only for getting random start value, in this case can we just generate random number from 0 to 1 and to get same chance to get needed item we can return something like

```
randomValue <= 0.1
  ? 4
  : 2;
```

wdyt?

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+  // if its impossible to create a new item:
+  } else {
+    document.querySelector('.message-lose').classList.remove('hidden');
+    document.querySelector('.message-play').classList.add('hidden');
+
+    if (Number(localStorage.getItem('score')) < Number(gameScore.innerText)) {
+      localStorage.setItem('score', gameScore.innerText);
+
+      highestScore.innerText = localStorage.getItem('score');
+    }
```
OleksiiBlezniuk, line 46: Mb let's move to helper function setNewHighestScore or smth like this?

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+  }
+}
+
+// Main movement algorithm:
+const arrowsPlayHandler = (e) => {
+  let isMoved = false;
+
+  if (e.key === 'ArrowDown') {
+    // MOVE DOWN:
+    // console.log('DOWN');
```
OleksiiBlezniuk, line 58: This comment can be removed, isn't it?)

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+          && fieldPos[1] === cellPos[1]
+          && cell.classList.length !== 1
+          && cell.id !== field.id
+          && field.innerText === cell.innerText
+          && checkBetween(fieldPos, cellPos, notEmptyFieldsPositions, 0)
+        ) {
+          margeCells(field, cell);
+          isMoved = true;
+          break;
+        }
```
OleksiiBlezniuk, line 87: we definitely can remove this `else` here

OleksiiBlezniuk: Better to somehow move this checks to helper functions, 'cause they are too big and it's hard to read

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+
+    document.body.removeEventListener('keyup', arrowsPlayHandler);
+  }
+}
+
+// Main movement algorithm:
+const arrowsPlayHandler = (e) => {
+  let isMoved = false;
+
+  if (e.key === 'ArrowDown') {
```
OleksiiBlezniuk, line 56: switch/case?

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+          && cell.classList.length !== 1
+          && cell.id !== field.id
+          && field.innerText === cell.innerText
+          && checkBetween(fieldPos, cellPos, notEmptyFieldsPositions, 0, true)
+        ) {
+          margeCells(field, cell);
+          isMoved = true;
+          break;
+        }
+      }
```
OleksiiBlezniuk, line 126: Let's move logic from this cases to separated functions, it's really hard to read

src/scripts/move.js
OleksiiBlezniuk: Also, can we make this logic reusable? I mean, seems that there is a kind of regularity in this checks and performed actions, better to separate this logic and make it reusable

OleksiiBlezniuk: For example pass to function values that had to be compared

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+          && checkBetween(fieldPos, cellPos, notEmptyFieldsPositions, 0)
+        ) {
+          margeCells(field, cell);
+          isMoved = true;
+          break;
+        }
+      }
+    }
+
+    if (isMoved || notEmptyFields.length === 16) {
```
OleksiiBlezniuk, line 91: Magic number `16`

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+      createItemInRandomEmptyField();
+    }
+  }
+};
+
+// function to marge 2 cells in one:
+function margeCells(curr, target) {
```
OleksiiBlezniuk, line 212:
```suggestion
function mergeCells(curr, target) {
```

src/scripts/move.js
```diff
@@ -0,0 +1,247 @@
...
+  }
+}
+
+// function to move a cell into the free field:
+function moveCell(curr, target) {
+  target.classList.add(`field-cell--${curr.innerText}`);
+  target.innerText = curr.innerText;
+  curr.classList.remove(`field-cell--${curr.innerText}`);
+  curr.innerText = '';
+}
```
OleksiiBlezniuk, line 242: Mb move to utils?

src/scripts/move.js
OleksiiBlezniuk: Mb move to utils?

OleksiiBlezniuk (changes requested):
Good job 🔥
Left some comments to fix, let's refactor your code to make it more understandable and readable

src/scripts/game.js
```diff
@@ -0,0 +1,64 @@
...
+      startBtn.classList.add(`restart`);
+
+      messageDuringTheGame.classList.remove('hidden');
+      messageBeforeStart.classList.add('hidden');
+
+      createItemInRandomEmptyField();
+      createItemInRandomEmptyField();
+
+      mobileSwipes(gameField);
+      document.body.addEventListener('keyup', arrowsPlayHandler);
```
OleksiiBlezniuk, line 42: I think this logic can be moved to separate method
I mean, I understand that it can't be reused, because it's a specific logic for one case, but this code inside if / else block is hard to read

It's not a critical issue, but would be nice to do it

src/scripts/move.js
```diff
@@ -0,0 +1,257 @@
...
+
+// Main movement algorithm:
+function arrowsPlayHandler(e) {
+  let isMoved = false;
+  const notEmptyFields = getNotEmptyFields(allCells);
+  const notEmptyFieldsPositions = getNotEmptyFieldsCoords(allCells);
+
+  switch (e.key) {
+    case 'ArrowDown':
+      for (const field of notEmptyFields.reverse()) {
```
OleksiiBlezniuk, line 69: Should we mutate this `notEmptyFields` array every time when this function works?

src/scripts/move.js
```diff
@@ -0,0 +1,257 @@
...
+        const fieldPos = field.id.split('-').slice(1);
+
+        for (const cell of [...allCells].reverse()) {
```
OleksiiBlezniuk, line 72: Well, it's exactly what I meant in previous comment =)
It's a bad practice to mutate incoming objects inside functions, 'cause it can be used somewhere  outside function
In case if there are no other options to do it, then ok, but better to avoid it

src/scripts/move.js
```diff
@@ -0,0 +1,257 @@
...
+    setNewHighestScore(gameScore, highestScore);
+
+    // mainListenerRemover();
+    document.body.removeEventListener('keyup', arrowsPlayHandler);
+  }
+}
+
+// Main movement algorithm:
+function arrowsPlayHandler(e) {
+  let isMoved = false;
```
OleksiiBlezniuk, line 63: analog of react state, nice 🙂

src/scripts/move.js
```diff
@@ -0,0 +1,257 @@
...
+          }
+        }
+      }
+
+      if (isMoved || notEmptyFields.length === totalCountOfCells) {
+        setTimeout(() => {
+          createItemInRandomEmptyField();
+        }, 301);
+      }
+      break;
```
OleksiiBlezniuk, line 128: But you've used the same logic for every case, but with other direction values (correct me if I wrong)
You can create separate function for it and reuse with different direction values 🙂

src/scripts/move.js
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// function to move a cell into the free field:
+function moveCell(curr, target, currCoords, targetCoords, axis) {
+  const transformAxis = axis === 0 ? 'Y' : 'X';
+  const value = curr.innerText;
+  const cssColor = getComputedStyle(curr).backgroundColor;
+
+  curr.style.transition = 'all 0.3s';
+  // curr.style.backgroundColor = moveCellColor;
```
OleksiiBlezniuk, line 235: it should be removed, isn't it? 🙂

OleksiiBlezniuk (changes requested):
You still can refactor `arrowsPlayHandler`, 'cause you still have duplicated logic 🙂

src/scripts/move.js
```diff
@@ -0,0 +1,230 @@
...
+        shouldReverse: false,
+      });
+      break;
+
+    case 'ArrowLeft':
+      mainMove({
+        mainAxis: 1,
+        staticAxis: 0,
+        revert: true,
+        shouldReverse: false,
```
OleksiiBlezniuk, line 95: Btw, you can skip this field, since it has false value, logic in `if` will work only if it has true value

src/scripts/move.js
```diff
@@ -0,0 +1,230 @@
...
+        )
+        && checkBetween(
+          fieldPos, cellPos, notEmptyFieldsPositions,
+          mainAxis, revert
+        )
+      ) {
+        mergeCells(field, cell, fieldPos, cellPos, mainAxis);
+        isMoved = true;
+        break;
+      }
```
OleksiiBlezniuk, line 153:
```suggestion
      const canMove = checkForMove(
        field, cell, fieldPos, cellPos,
        mainAxis, staticAxis, revert
      );
      const canMerge = checkForMerge(
        field, cell, fieldPos, cellPos,
        mainAxis, staticAxis, revert
      );
      const isBetween = checkBetween(
        fieldPos, cellPos, notEmptyFieldsPositions,
        mainAxis, revert
      );

      if (canMove) {
        moveCell(field, cell, fieldPos, cellPos, mainAxis);
        isMoved = true;
        break;
      } 
      
      if (canMerge && isBetween)  { // since you've used break before
        mergeCells(field, cell, fieldPos, cellPos, mainAxis);
        isMoved = true;
        break;
      }
```

OleksiiBlezniuk: Let's move this checks result to variable, it's will be much easier to understand conditions

OleksiiBlezniuk (approved):
Looks much better then at first implementation, well done!

PR #113, opened 2022-09-26
first commit, not finished

vitalii-heiko:
[DEMO LINK](https://vitalii-heiko.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,118 @@
 'use strict';
 
 // write your code here
+const button = document.querySelector('.button');
+const messeageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const gameScore = document.querySelector('.game-score');
+const rows = document.querySelectorAll('.field-row');
+const rowCells = 4;
+const cells = 4;
+let score = 0;
+let gameField = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
+
+function setGameField() {
+  for (let row = 0; row < rowCells; row++) {
+    for (let cell = 0; cell < cells; cell++) {
+      rows[row].children[cell].textContent = gameField[row][cell] || '';
+
+      rows[row].children[cell].className
+        = `field-cell field-cell--${gameField[row][cell]}`;
+
+      if (gameField[row][cell] === 2048) {
+        messageWin.classList.remove('hidden');
+      };
+    }
+  }
+
+  // checkGameOver();
+};
+
+function hasEmptyCell() {
+  for (let row = 0; row < rowCells; row++) {
+    for (let cell = 0; cell < cells; cell++) {
+      if (gameField[row][cell] === 0) {
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+function setTwoOrFour() {
+  if (!hasEmptyCell()) {
+    messageLose.classList.remove('hidden');
+
+    return;
+  }
+
+  const row = Math.floor(Math.random() * rowCells);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
 'use strict';
 
 // write your code here
+const button = document.querySelector('.button');
+const messeageStart = document.querySelector('.message-start');
```
qoa11a, line 5: Typo.
```suggestion
const messageStart = document.querySelector('.message-start');
```

src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+function canMove() {
+  for (let r = 0; r < 4; r++) {
```
qoa11a, line 46: It's better to write full word instead of one letter as it can be unclear. Moreover, you've already wrote two functions with full words and it's good to keep same logic for variables with same meaning.
```suggestion
  for (let row = 0; row < 4; row++) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+    for (let c = 0; c < 3; c++) {
```
qoa11a, line 47:
```suggestion
    for (let cell = 0; cell < 3; cell++) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+function setTwoOrFour() {
+  canMove();
```
qoa11a, line 59: Why do you call it here?

src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+    return;
+  }
+
+  const row = Math.floor(Math.random() * 4);
+  const cell = Math.floor(Math.random() * 4);
+
+  if (hasEmptyCell()) {
+    (gameField[row][cell] === 0)
+      ? gameField[row][cell] = Math.floor(Math.random() >= 0.9 ? 4 : 2)
+      : setTwoOrFour();
```
qoa11a, line 73: It's better to avoid recursion here as you repeat check for the ability to move and presence of empty cells even though the game field hasn't changed. You should only redeclare random row and cell for them to be in empty field.

src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+let gameField = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
+
+function setGameField() {
+  for (let row = 0; row < 4; row++) {
+    for (let cell = 0; cell < 4; cell++) {
```
qoa11a, line 20: Why do you call it cell? It's more logical to call it column as it represents the column. You should change it all over the project.

src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+  }
+
+  setGameField();
+}
+
+function filteringZero(row) {
+  return row.filter(num => num !== 0);
+}
+
+function slide(row) {
```
qoa11a, line 83: In can be a column so the argument name is unclear.

src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+
+button.addEventListener('click', () => {
+  if (button.textContent === 'Start') {
+    button.textContent = 'Restart';
+    button.classList.replace('start', 'restart');
+    messeageStart.classList.add('hidden');
+    messageWin.classList.add('hidden');
+    messageLose.classList.add('hidden');
+    setTwoOrFour();
+    setTwoOrFour();
```
qoa11a, line 172: I recommend you to move this to separate functions like startGame() and restartGame().

qoa11a (changes requested):

Bohdan-Maksymiuk (approved):

PR #115, opened 2022-09-28
Develop

grafit-off:
You also can use mouse moves for sliding (like Drag'n'Drop) instead of arrow keys.
 - [DEMO LINK](https://grafit-off.github.io/js_2048_game/)


MaxManis (approved):
looks complicated but its works good!)

silchyn (approved):
Cool!

src/index.html
```diff
@@ -1,63 +1,87 @@
...
-        <p class="message message-lose hidden">You lose! Restart the game?</p>
-        <p class="message message-win hidden">Winner! Congrats! You did it!</p>
-        <p class="message message-start">Press "Start" to begin game. Good luck!</p>
-      </div>
+      <footer class="game__footer footer">
+        <!--
+          <p class="message message-lose hidden">You lose! Restart the game?</p>
+          <p class="message message-win hidden">Winner! Congrats! You did it!
+          </p>
+          -->
```
OleksiiBlezniuk, line 74: Why do you need this commented paragraphs?

src/scripts/main.js
```diff
@@ -1,3 +1,446 @@
...
+  }
+
+  getColumnLine(columnIndex) {
+    return [
+      this.fieldState[0][columnIndex],
+      this.fieldState[1][columnIndex],
+      this.fieldState[2][columnIndex],
+      this.fieldState[3][columnIndex],
+    ];
+  }
```
OleksiiBlezniuk, line 404: Probably in order to make this game field size changable (if we'll want to implement this feature in future) would be better to make something like loop from 0 to `fieldSize - 1` instead of hardcoding, but now it's ok

src/scripts/main.js
```diff
@@ -1,3 +1,446 @@
...
+
+    let emptyTileFound = false;
+
+    while (!emptyTileFound) {
+      const [rowIndex, columnIndex] = this.generateRandomTilePosition();
+
+      if (this.fieldState[rowIndex][columnIndex] === 0) {
+        const tile = this.getTileElementByPosition(rowIndex, columnIndex);
+        const tileValue = this.generateNewTileValue();
+        const { keyframes, options } = this.tileAppearanceAnimation;
```
OleksiiBlezniuk, line 327: 🔥

OleksiiBlezniuk (approved):
This class looks nice 🔥 
If you want to improve this game, I can suggest some improvements:
- try to add some transition to tiles move
- add something like highestScore and add it's value to localStorage
- save gameFields state to localStorage

But still, it's a nice-to-have features, task is already completed, well done! 🚀

PR #122, opened 2022-10-16
PC with keys, PC with mouse, mobile with touch

TheVovchik:
[DEMO LINK](https://TheVovchik.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,344 @@
 'use strict';
 
-// write your code here
+class Game {
+  constructor(game) {
+    this.fieldRows = document.querySelectorAll('.field-row');
+    this.game = game;
+    this.startButton = document.querySelector('.start');
+    this.highScoreValue = 0;
+    this.bestScore = document.querySelector('.best-score');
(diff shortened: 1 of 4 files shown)


silchyn (changes requested):
Restart does not reset the score. Also, the score and best score can occasionally change when constantly restarting

silchyn (changes requested):
Has something changed?

src/scripts/main.js
```diff
@@ -1,3 +1,435 @@
...
+    if (e.key === 'ArrowUp') {
+      this.verticalUp();
+    } else if (e.key === 'ArrowDown') {
+      this.verticalDown();
+    } else if (e.key === 'ArrowLeft') {
+      this.horizontalLeft();
+    } else if (e.key === 'ArrowRight') {
+      this.horizontalRight();
+    } else {
+      return;
```
Anna-Tkachenko, line 178: use switch / case if you need to call some functionality depending on the value of the one variable

Anna-Tkachenko (approved):
Code is clean and detailed, good job

PR #124, opened 2022-10-28
develop

YaroslavKolbaiev:
[DEMO LINK](https://YaroslavKolbaiev.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,260 @@
 'use strict';
 
-// write your code here
+let board;
+const rows = 4;
+const columns = 4;
+const score = document.querySelector('.game-score');
+const cells = document.querySelectorAll('.field-cell');
+const button = document.querySelector('.button');
+
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const messageStart = document.querySelector('.message-start');
+
+button.addEventListener('click', () => {
+  if (button.textContent === 'Start') {
+    button.classList.remove('start');
+    button.classList.add('restart');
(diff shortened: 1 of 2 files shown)


polosanya (changes requested):
It's impossible to lose now. Please fix it
image

loralevitska (changes requested):
Your game is fun. Good job!
Let's improve it. Fix it

image

src/scripts/main.js
```diff
@@ -1,3 +1,258 @@
...
+      const tile = document.getElementById(`${r}-${c}`);
+      const num = board[r][c];
+
+      updateTile(tile, num);
+    }
+  }
+}
+
+function slideUp() {
+  for (let c = 0; c < columns; c++) {
```
loralevitska, line 200: use meaningful variables - `column` `index` etc

src/scripts/main.js
```diff
@@ -1,3 +1,258 @@
...
+
+      const tile = document.getElementById(`${r}-${c}`);
+      const num = board[r][c];
+
+      updateTile(tile, num);
+    }
+  }
+}
+
+function slideDown() {
```
loralevitska, line 216: there are many similarities in these functions `slideDown` `slideUp`. Merge into one by passing different parameters to avoid duplicate code or leave these two functions, and in the third auxiliary take out everything that is the same in these two

loralevitska (changes requested):
amazing :star-struck:

try to write more meaningful name of commit. E.g. `fix: losing option`
image

Anastasiia-Svintsova (approved):
Good job!

PR #128, opened 2022-11-07
solution

TetianaKoval:
[DEMO LINK](https://TetianaKoval.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,340 @@
 'use strict';
 
-// write your code here
+const cells = document.querySelectorAll('.field-cell');
+const table = document.querySelector('.game-field');
+const start = document.querySelector('.start');
+const messegeStart = document.querySelector('.message-start');
+const messegeLose = document.querySelector('.message-lose');
+const messegeWin = document.querySelector('.message-win');
+let score = 0;
+const scoreInGame = document.querySelector('.game-score');
+
+function createNumber() {
+  const emptyCells
+    = [...document.querySelectorAll('.field-cell')]
+      .filter(cell => !cell.innerText);
+
+  function random(minNumber, maxNumber) {
+    const min = Math.ceil(minNumber);
+    const max = Math.floor(maxNumber);
+
+    return Math.floor(Math.random()
+      * (max - min + 1)) + min; // Максимум и минимум включаются
+  };
+
+  const randomCellIndex = random(0, [...emptyCells].length - 1);
+  let number;
+  const randomNumber = random(0, 100);
+
+  if (randomNumber >= 90) {
+    number = 4;
+  } else {
+    number = 2;
+  };
+
+  function addNumber(cell) {
+    cell.setAttribute('id', 'new-number');
+
+    setTimeout(() => {
+      cell.removeAttribute('id');
+    }, 100);
+
+    cell.innerText = number;
+  };
+  addNumber(emptyCells[randomCellIndex]);
+
+  addClass([...cells]);
+};
+
+function addClass(cellsArr) {
+  cellsArr.map(cell => {
+    if (cell.innerText) {
+      cell.className = `field-cell field-cell--${cell.innerText}`;
+    } else {
+      cell.className = ``;
+      cell.classList.add(`field-cell`);
+    }
+  });
+};
+
+const tbodyHorisontal = document.querySelector('tbody');
+const tbodyVertical = document.createElement('tbody');
+
+tbodyVertical.classList.add('tbody-columns');
+
+// розворот таблиці
+function turnTheTable(tableBody, type) {
+  tableBody.innerHTML = '';
+
+  for (let i = 0; i < document.querySelector('tbody').children.length; i++) {
+    // отримую колонку
+    const cellsInColumn = [...document.querySelectorAll('td')]
+      .filter(cell => cell.cellIndex === i);
+    // створюю таблицю де стовпці це рядки
+    const newLine = document.createElement('tr');
+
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+const messegeStart = document.querySelector('.message-start');
+const messegeLose = document.querySelector('.message-lose');
+const messegeWin = document.querySelector('.message-win');
+let score = 0;
+const scoreInGame = document.querySelector('.game-score');
+
+function createNumber() {
+  const emptyCells
+    = [...document.querySelectorAll('.field-cell')]
+      .filter(cell => !cell.innerText);
```
qoa11a, line 15:
```suggestion
  const emptyCells = [...document.querySelectorAll('.field-cell')]
    .filter(cell => !cell.innerText);
```

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+
+  function random(minNumber, maxNumber) {
+    const min = Math.ceil(minNumber);
+    const max = Math.floor(maxNumber);
+
+    return Math.floor(Math.random()
+      * (max - min + 1)) + min; // Максимум и минимум включаются
+  };
```
qoa11a, line 23: Move helper functions outside other functions.

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+  const emptyCells
+    = [...document.querySelectorAll('.field-cell')]
+      .filter(cell => !cell.innerText);
+
+  function random(minNumber, maxNumber) {
+    const min = Math.ceil(minNumber);
+    const max = Math.floor(maxNumber);
+
+    return Math.floor(Math.random()
+      * (max - min + 1)) + min; // Максимум и минимум включаются
```
qoa11a, line 22:
```suggestion
    return Math.floor(
      Math.random() * (max - min + 1)
    ) + min; // Максимум и минимум включаются
```

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+          possibleNextStep++;
+        }
+      }
+    };
+
+    if (possibleNextStep === 0) {
+      messegeLose.classList.remove('hidden');
+      messegeStart.classList.add('hidden');
+      messegeWin.classList.add('hidden');
+      document.removeEventListener('keydown', keydownListener);
```
qoa11a, line 311: Move such logic to a function with appropriate name. Overall, I recommend to move a lot of conditions inside of "if" statements, some logic that can be easily described as a function name etc., to separate functions. Then code becomes easily readable and is easy to support.
https://refactoring.guru/extract-method

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+    }
+  };
+}
+
+// Кнопка start
+start.addEventListener('click', (e) => {
+  document.addEventListener('keydown', keydownListener);
+  messegeStart.classList.add('hidden');
+  messegeLose.classList.add('hidden');
+  messegeWin.classList.add('hidden');
```
qoa11a, line 321: I see these methods not only here. DRY - don't repeat yourself.
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+};
+
+// перемога
+function victory(valuesInCells) {
+  if (valuesInCells.find(cell => cell === '2048')) {
+    messegeStart.classList.add('hidden');
+    messegeLose.classList.add('hidden');
+    messegeWin.classList.remove('hidden');
+
+    document.removeEventListener('keydown', keydownListener);
```
qoa11a, line 243: Repetative code may be found below in code. DRY.

qoa11a (changes requested):
Great job! 👍 

Read comments.
I want to recommend you to write less comments. It's better when the [code is self-documenting](https://en.wikipedia.org/wiki/Self-documenting_code). Try to extract logic and conditions to functions with appropriate names, than there won't be need in comments.

Another point is that you should don't repeat yourself. It's a principle that basically means that if you write something twice or more times than you should move it to a function or a variable. You repeat some parts of code three or more times.

And notice how hard to read are some "if" statements. I recommend to name these conditions and move them to variables or functions. Than just use them inside "if" statements.

src/scripts/main.js
```diff
@@ -1,3 +1,355 @@
...
+    insertTable('vetical', horisontalTable);
+    rowsRewriteToRight();
+  };
+
+  if (e.key === 'ArrowLeft') {
+    insertTable('horisontal', horisontalTable);
+    rowsRewriteToLeft();
+  };
+
+  if (e.key === 'ArrowRight') {
```
polosanya, line 227: It is better to use switch...case instead
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

src/scripts/main.js
```diff
@@ -1,3 +1,355 @@
...
+
+        if (!verticalNumbers[allCells[i].parentElement.rowIndex]) {
+          verticalNumbers[allCells[i].parentElement.rowIndex] = [];
+        };
+
+        verticalNumbers[allCells[i].parentElement.rowIndex]
+          .push(allCells[i].innerText);
+      };
+    } else {
+      for (let i = 0; i < [...allCells].length; i++) {
```
polosanya, line 278: In both cases you use the same loop with similar logic. 
Try to simplify your code, and use only one loop

polosanya (changes requested):
Nice! Just have some comments that may help you improve your code

qoa11a (approved):

PR #131, opened 2022-11-09
Solution 2048 game

volodymir-tymtsias:
[DEMO LINK](https://volodymir-tymtsias.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,246 @@
 'use strict';
 
-// write your code here
+const tableBody = document.querySelector('.game-field').firstElementChild;
+const buttonStart = document.querySelector('.start');
+const gameScore = document.querySelector('.game-score');
+const rows = [...tableBody.rows];
+const cells = [];
+let filledCells;
+let epmtyCells;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,246 @@
 'use strict';
 
-// write your code here
+const tableBody = document.querySelector('.game-field').firstElementChild;
```
TarasHoliuk, line 3:
```suggestion
const tableBody = document.querySelector('tbody');
```

src/scripts/main.js
```diff
@@ -1,3 +1,246 @@
...
+const buttonStart = document.querySelector('.start');
+const gameScore = document.querySelector('.game-score');
+const rows = [...tableBody.rows];
```
TarasHoliuk, line 6:
```suggestion
const rows = document.querySelectorAll('.field-row');
```
And then no need to query 'tbody'

TarasHoliuk (approved):
GJ! I've added just a few minor comments about `querySelectors` check it, please

PR #132, opened 2022-11-10
solution

serhii-yunnikov:
https://serhii-yunnikov.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,199 @@
 'use strict';
 
-// write your code here
+const cells = [...document.querySelectorAll('.field-cell')];
+const start = document.querySelector('.start');
+const rowsCollection = document.querySelectorAll('.field-row');
+const score = document.querySelector('.game-score');
+const messages = document.querySelectorAll('.message');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+
+cells.forEach((cell, index) => {
+  cell.dataset.num = `${index + 1}`;
+});
+
+function generateEven() {
+  const num = Math.random();
+
+  return num <= 0.1 ? 4 : 2;
+}
+
+function generateCellNoRepeated() {
+  const randomArr = [];
+  const number1 = Math.ceil(Math.random() * 16);
+
+  randomArr.push(number1);
+
+  let number2 = Math.ceil(Math.random() * 16);
+
+  while (randomArr.includes(number2)) {
+    number2 = Math.ceil(Math.random() * 16);
+  }
+
+  randomArr.push(number2);
+
+  return randomArr;
+}
+
+function erase() {
+  cells.forEach(cell => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+    leftCellsNum();
+    moveCells();
+    mergeCells();
+  }
+
+  if (e.code === 'ArrowRight') {
+    rightCellsNum();
+    moveCells();
+    mergeCells();
+  }
```
Anastasiia-Svintsova, line 183: use switch/case here

src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+
+  freeCell.innerText = `${freeCellValue}`;
+  freeCell.classList.add(`field-cell--${freeCellValue}`);
+}
+
+function mergeCells() {
+  for (let n = 1; n <= 12; n++) {
+    const valueUp = cellNo(n).innerText;
+
+    if (cellNo(n).className === cellNo(n + 4).className
```
Anastasiia-Svintsova, line 92: you use `cellNo(n)` and `cellNo(n + 4)` so many times, maybe it would be better to create variables for them

src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+
+function generateEven() {
+  const num = Math.random();
+
+  return num <= 0.1 ? 4 : 2;
+}
+
+function generateCellNoRepeated() {
+  const randomArr = [];
+  const number1 = Math.ceil(Math.random() * 16);
```
Anastasiia-Svintsova, line 23: move `Math.ceil(Math.random() * 16)` to a separate method - `generateRandomNumber()` or smth like this

Anastasiia-Svintsova (changes requested):
Good! Just fix some comments

qoa11a (changes requested):
image
I've lost even though I have possible move (up). Fix it please.

src/scripts/main.js
```diff
@@ -1,3 +1,222 @@
...
+      leftCellsNum();
+      moveCells();
+      mergeCells();
+      break;
+    case 'ArrowRight':
+      rightCellsNum();
+      moveCells();
+      mergeCells();
+      break;
+  }
```
Anastasiia-Svintsova, line 206: It's hard to read
```suggestion
  switch (e.code) {
    case 'ArrowDown':
      downCellsNum();
      moveCells();
      mergeCells();
      break;
      
    case 'ArrowUp':
      upCellsNum();
      moveCells();
      mergeCells();
      break;
      
    case 'ArrowLeft':
      leftCellsNum();
      moveCells();
      mergeCells();
      break;
      
    case 'ArrowRight':
      rightCellsNum();
      moveCells();
      mergeCells();
      break;
  }
```

Anastasiia-Svintsova (approved):

PR #133, opened 2022-11-10
Develop

oleksii-bieliaiev:
[DEMO LINK](https://oleksii-bieliaiev.github.io/js_2048_game/)

Use the arrow keys on your keyboard to play: up, down, left, and right.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -130,12 +130,12 @@ function checkConnection() {
 
 function moveRight() {
   for (let row = 0; row < width; row++) {
-    let el = field[row].reverse();
+    const el = field[row].reverse();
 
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,244 @@
 'use strict';
 
-// write your code here
+let field = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
```
qoa11a, line 8: You can move this initial state into variable as you use it twice.

qoa11a (approved):
Good job! 👍

PR #136, opened 2022-11-11
2048 game was created

illia-kots:
[DEMO LINK](https://illia-kots.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,204 @@
 'use strict';
 
-// write your code here
+const buttonStart = document.querySelector('.start');
+const rowField = document.querySelectorAll('.field-row');
+const gameScore = document.querySelector('.game-score');
+const messageStart = document.querySelector('.message-start');
+const messageRestart = document.querySelector('.message-restart');
+const messageWin = document.querySelector('.message-win');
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,204 @@
...
+
+  gameStart();
+});
+
+document.addEventListener('keyup', (key) => {
+  if (buttonStart.classList.contains('restart')) {
+    if (key.code === 'ArrowRight') {
+      moveMix(moveRight, combineRow);
+      randomize();
+    } else if (key.code === 'ArrowLeft') {
```
polosanya, line 158: It is better to use switch...case instead
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

polosanya (changes requested):
Looks great! Please pay attention to my comment

src/scripts/main.js
illia-kots: Ok. I'll change to "switch case". Thanks for recommendation!)

polosanya (approved):
Great! I'm approving your job, but please pay attention to my comments

PR #139, opened 2022-11-14
game 2048 is done

ruslik87:


src/scripts/main.js
```diff
@@ -1,3 +1,226 @@
...
+      shift('up');
+      checkEndOfGame();
+      generate();
+      break;
+
+    case 'ArrowDown':
+      shift('down');
+      checkEndOfGame();
+      generate();
+      break;
```
Bohdan-Maksymiuk, line 58: looks like duplicated ligic - try to unite into separate function

src/scripts/main.js
```diff
@@ -1,3 +1,226 @@
...
+
+  for (const row in cellsCoords) {
+    if (cellsCoords[row].includes(0)) {
+      lose = false;
+    }
+
+    for (let i = 0; i < cellsCoords[row].length - 1; i++) {
+      if (
+        cellsCoords[row][i] === cellsCoords[row][i + 1]
+        || cellsCoords[i][row] === cellsCoords[i + 1][row]
```
Bohdan-Maksymiuk, line 216: pass such large conditions into separate variable

Bohdan-Maksymiuk (changes requested):
* better to `start` game after `restart` automatically 

Good job, lets make a little improvements of code

Bohdan-Maksymiuk (approved):

DarkMistyRoom (changes requested):
Great! But look how it works in the reference. High score updates on each move when the score is higher than the previous value. And it is stored after the "new game"

marina-tilniak (approved):

PR #144, opened 2022-11-23
2048

olenitut:
- [DEMO LINK](https://olenitut.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,257 @@
 'use strict';
 
-// write your code here
+/// /// /// /// /// /// ///
+//    APP LOGIC
+/// /// /// /// /// /// ///
+
+let score = 0;
+let isPlaying = false;
+
+const row1 = [0, 1, 2, 3];
+const row2 = [4, 5, 6, 7];
+const row3 = [8, 9, 10, 11];
+const row4 = [12, 13, 14, 15];
+
+const column1 = [0, 4, 8, 12];
+const column2 = [1, 5, 9, 13];
+const column3 = [2, 6, 10, 14];
+const column4 = [3, 7, 11, 15];
+
+const rightDirection = [column3, column2, column1].flat();
+const leftDirection = [column2, column3, column4].flat();
+const upDirection = [row2, row3, row4].flat();
+const downDirection = [row3, row2, row1].flat();
+
+/// /// /// /// /// /// ///
+/// ELEMENTS SELECTION AND CRIATION
+/// /// /// /// /// /// ///
+
+const cells = document.querySelectorAll('.field-cell');
+const startButton = document.querySelector('.start');
+const restartButton = document.createElement('button');
+const messageLoseElement = document.querySelector('.message-lose');
+const messageWinElement = document.querySelector('.message-win');
+const messageStartElement = document.querySelector('.message-start');
+
+restartButton.classList.add('button', 'restart');
+restartButton.textContent = 'Restart';
+restartButton.hidden = true;
+document.querySelector('.controls').append(restartButton);
+
+/// /// /// /// /// /// ///
+/// HELPER FUNCTIONS
+/// /// /// /// /// /// ///
+const checkIfHaveMoves = () => {
+  if ([...cells].filter(cell => cell.classList.length === 1).length > 0) {
+    return true;
+  }
+
+  let haveMoves = false;
+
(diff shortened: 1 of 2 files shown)


liliia-livandovskaya (approved):
Awesome! 🔥

src/scripts/main.js
```diff
@@ -1,3 +1,257 @@
...
+restartButton.classList.add('button', 'restart');
+restartButton.textContent = 'Restart';
+restartButton.hidden = true;
+document.querySelector('.controls').append(restartButton);
+
+/// /// /// /// /// /// ///
+/// HELPER FUNCTIONS
+/// /// /// /// /// /// ///
+const checkIfHaveMoves = () => {
+  if ([...cells].filter(cell => cell.classList.length === 1).length > 0) {
```
OlEzhA0, line 45: why do we need copy array for filter method?

src/scripts/main.js
```diff
@@ -1,3 +1,257 @@
...
+
+    const adjacentCells = [topCell, bottomCell, leftCell, rightCell];
+
+    adjacentCells.forEach(adjacentCell => {
+      if (!adjacentCell) {
+        return;
+      }
+
+      if (adjacentCell.textContent === cell.textContent) {
+        haveMoves = true;
```
OlEzhA0, line 65: add `return` here; if we know that we have moves, we may not check other elements

src/scripts/main.js
```diff
@@ -1,3 +1,257 @@
...
+    }
+  }
+
+  return cells[cellIndex];
+};
+
+const generateNumberForTile = () => {
+  const num = Math.random();
+
+  return num < 0.9 ? 2 : 4;
```
OlEzhA0, line 100: in original 2048 `8` can also be generated

src/scripts/main.js
```diff
@@ -1,3 +1,257 @@
...
+  if (nextCell.classList.length > 1) {
+    return;
+  };
+
+  updateCell(currentCell);
+  updateCell(nextCell, cellValue);
+};
+
+const addHelper = (currentCell, nextCell, cellValue) => {
+  if (Number(nextCell.textContent) === Number(cellValue)) {
```
OlEzhA0, line 126: this syntax more preferable 
```suggestion
  if (+nextCell.textContent === +cellValue) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,257 @@
...
+    case 'left':
+      step = -1;
+      break;
+    case 'up':
+      step = -4;
+      break;
+    case 'down':
+      step = 4;
+      break;
+  };
```
OlEzhA0, line 154: you can use next trick
```
const steps = {
  right: 1,
  left: -1,
  up: -4,
  down: 4
};

const step = steps[direction];
```

src/scripts/main.js
```diff
@@ -1,3 +1,257 @@
...
+const moveDirectionAndSum = (directionArr, direction) => {
+  moveAddTiles(directionArr, direction);
+  moveAddTiles(directionArr, direction);
+  moveAddTiles(directionArr, direction);
+  moveAddTiles(directionArr, direction, 'add');
+  moveAddTiles(directionArr, direction);
+  moveAddTiles(directionArr, direction);
+};
+
+const innit = () => {
```
OlEzhA0, line 184:
```suggestion
const init = () => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,257 @@
...
+
+const documentCallbak = (e) => {
+  e.preventDefault();
+
+  if (!checkIfHaveMoves()) {
+    isPlaying = false;
+    messageLoseElement.classList.remove('hidden');
+  };
+
+  const currentCellsValues = [...cells].map(cell => cell.textContent);
```
OlEzhA0, line 220: why do we need copy array for map method?

src/scripts/main.js
```diff
@@ -1,3 +1,257 @@
...
+    case 'ArrowRight':
+      moveDirectionAndSum(rightDirection, 'right');
+      break;
+  };
+
+  if (!isPlaying) {
+    return;
+  }
+
+  const newCellsValues = [...cells].map(cell => cell.textContent);
```
OlEzhA0, line 245: the same

OlEzhA0 (approved):

PR #151, opened 2022-12-11
solution

1stIndominus:
[DEMO LINK](https://1stIndominus.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,3 +1,237 @@
...
+
+      updateTile(tile, num);
+    }
+  }
+};
+
+function lose() {
+  if (hasEmptyTile()) {
+    return;
+  }
```
loralevitska, line 216: This check better use outside from function. From your function name is not obvious what it does. It really has conditions and if all is well, it checks the value and makes some text visible but not lose

src/scripts/main.js
```diff
@@ -1,3 +1,237 @@
...
+      if (board[c][r] === board[c + 1][r]
+        || board[r][c] === board[r][c + 1]) {
+        loseMessage.classList.remove('hidden');
+        document.removeEventListener('keyup', handler);
+      }
+    }
+  }
+}
+
+function win() {
```
loralevitska, line 229: Same here, the name of the function does not reproduce the validity

loralevitska (approved):
good job!

PR #163, opened 2022-12-27
2048 game has been created

miamarichka:
[DEMO LINK] https://miamarichka.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,193 @@
 'use strict';
 
-// write your code here
+const table = document.querySelector('table');
+let score = 0;
+
+let board = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
+
+const rows = 4;
+const columns = 4;
+
+const button = document.querySelector('button');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+const messageStart = document.querySelector('.message-start');
+
+button.addEventListener('click', () => {
+  if (button.classList.contains('start')) {
+    button.classList.remove('start');
+    button.classList.add('restart');
+    button.innerText = 'Restart';
+
+    messageStart.classList.add('hidden');
+    setGame();
+    setGame();
+    updateGame();
+  } else {
+    if (button.classList.contains('restart')) {
+      button.classList.remove('restart');
+      button.classList.add('start');
+      messageStart.classList.remove('hidden');
+      button.innerText = 'Start';
+
+      board = [
+        [0, 0, 0, 0],
+        [0, 0, 0, 0],
+        [0, 0, 0, 0],
+        [0, 0, 0, 0],
+      ];
+
+      messageLose.classList.add('hidden');
+      messageWin.classList.add('hidden');
+      score = 0;
+
+      updateGame();
+    }
+  };
+});
+
+function setGame() {
+  const emptyTiles = [];
+
+  board.forEach((row, rowIndex) => {
+    row.forEach((cell, cellIndex) => {
+      if (cell === 0) {
+        emptyTiles.push([rowIndex, cellIndex]);
+      }
+    });
+  });
+
+  if (!emptyTiles.length) {
+    return;
+  }
+
+  const randomizer = (Math.random() <= 0.1) ? 4 : 2;
+  const [rowI, cellI]
+        = emptyTiles[Math.floor(Math.random() * emptyTiles.length)];
+
+  board[rowI][cellI] = randomizer;
+}
+
+function isGameOver() {
+  const anyEmptyField = board.some(row => row.some(cell => cell === 0));
+
+  if (anyEmptyField) {
+    return;
+  }
+
+  for (let r = 0; r < rows; r++) {
+    for (let c = 1; c < columns; c++) {
+      if (board[r][c] === board[r][c + 1]
+                || board[c][r] === board[c + 1][r]) {
+        return;
+      }
+    }
+  }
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,193 @@
...
+
+button.addEventListener('click', () => {
+  if (button.classList.contains('start')) {
+    button.classList.remove('start');
+    button.classList.add('restart');
+    button.innerText = 'Restart';
+
+    messageStart.classList.add('hidden');
+    setGame();
+    setGame();
```
Anastasiia-Svintsova, line 29: Do you need to call this function twice?

src/scripts/main.js
```diff
@@ -1,3 +1,193 @@
...
+      button.classList.add('start');
+      messageStart.classList.remove('hidden');
+      button.innerText = 'Start';
+
+      board = [
+        [0, 0, 0, 0],
+        [0, 0, 0, 0],
+        [0, 0, 0, 0],
+        [0, 0, 0, 0],
+      ];
```
Anastasiia-Svintsova, line 43: Create a variable for a default board
```
const defaultBoard = new Array(4).fill([0, 0, 0, 0]);
let board = defaultBoard;

File: src/scripts/main.js [hunk 3]
```diff
@@ -1,3 +1,193 @@
...
+
+      messageLose.classList.add('hidden');
+      messageWin.classList.add('hidden');
+      score = 0;
+
+      updateGame();
```
Anastasiia-Svintsova, line 49: You call  `updateGame()` in both cases. So just call it at the end of the function once

src/scripts/main.js
```diff
@@ -1,3 +1,193 @@
...
+    button.classList.remove('start');
+    button.classList.add('restart');
+    button.innerText = 'Restart';
+
+    messageStart.classList.add('hidden');
+    setGame();
+    setGame();
+    updateGame();
+  } else {
+    if (button.classList.contains('restart')) {
```
Anastasiia-Svintsova, line 32:
```suggestion
  } else if (button.classList.contains('restart')) {
```

Anastasiia-Svintsova (changes requested):
- Can't see a message that the game is over
image
- The game should start after pressing on the restart button

https://user-images.githubusercontent.com/99403198/209805805-fc7e4aa7-8192-4fa6-a20d-28c8adfffa30.mov

src/scripts/main.js
miamarichka: > Do you need to call this function twice?

yes, because its generate one tile, but I need 2

src/scripts/main.js
miamarichka: I did as you your example, but its didnt work, I tried to find out, but coudnt. So I used defaultBoard value for board twice

src/scripts/main.js
miamarichka: I`ve made some changes, so now -only once

src/scripts/main.js
```diff
@@ -1,3 +1,176 @@
...
+
+        for (let i = 0; i < rows; i++) {
+          board[i][c] = row[i];
+        };
+      }
+      break;
+  }
+
+  setGame();
+  updateGame();
```
loralevitska, line 175: Is it really necessary to do this right here, and not when clicking on a button?

src/scripts/main.js
```diff
@@ -1,3 +1,176 @@
...
+        row = slide(row);
+        row.reverse();
+        board[0][c] = row[0];
+        board[1][c] = row[1];
+        board[2][c] = row[2];
+        board[3][c] = row[3];
+
+        for (let i = 0; i < rows; i++) {
+          board[i][c] = row[i];
+        };
```
loralevitska, line 169: Repeatable code move to helper functions

loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements (check comments)

1. I have 8 filled cells since the beginning of the game. Need 2
image
2. After restarting the game, there should be 2 filled fields
image

src/scripts/main.js
miamarichka: I did it 4 different functions for every movement

src/scripts/main.js
miamarichka: Yes, because when I do click - I move it and after any of 4 movement I update it, in another case I need to call it for every movement, so 4 times for every

src/scripts/main.js
```diff
@@ -1,3 +1,197 @@
...
+
+button.addEventListener('click', () => {
+  if (button.classList.contains('start')) {
+    button.classList.remove('start');
+    button.classList.add('restart');
+    button.innerText = 'Restart';
+
+    messageStart.classList.add('hidden');
+  } else if (button.classList.contains('restart')) {
+    board = new Array(4).fill([0, 0, 0, 0]);
```
Anastasiia-Svintsova, line 26: You forgot to fix this value and after restarting, I can see 8 field cells instead of 2

src/scripts/main.js
```diff
@@ -1,3 +1,197 @@
...
+    board[0][c] = row[0];
+    board[1][c] = row[1];
+    board[2][c] = row[2];
+    board[3][c] = row[3];
+
+    for (let i = 0; i < rows; i++) {
+      board[i][c] = row[i];
+    }
+  }
+};
```
Anastasiia-Svintsova, line 197: These functions are very similar to each other. It may be worth combining them into one `function verticalMove`, passing `up` or `down` in the argument. If `down`, then do additional actions

Anastasiia-Svintsova (changes requested):

SanyaBratashchuk (approved):

PR #164, opened 2022-12-29
Solution

ifuride:
[DEMO LINK](https://ifuride.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,338 @@
 'use strict';
 
-// write your code here
+const rowLength = 4;
+const columnLength = 4;
+let score = 0;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,338 @@
...
+      mergeRight(field);
+      moveRight(field);
+      break;
+
+    case 'ArrowLeft':
+      moveLeft(field);
+      mergeLeft(field);
+      moveLeft(field);
+      break;
+  }
```
polosanya, line 319: Don't forget to add a default case to handle situations when key doesn't match any of cases

polosanya (approved):

PR #173, opened 2023-01-11
js_2048_game

rafmamedov:
- [DEMO LINK](https://rafmamedov.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,258 @@
 'use strict';
 
-// write your code here
+let gameField;
+let score = 0;
+const rows = 4;
+const columns = 4;
+const field = document.querySelector('tbody');
+const button = document.querySelector('.button');
+const startMessage = document.querySelector('.message-start');
+const loseMessage = document.querySelector('.message-lose');
+const winMessage = document.querySelector('.message-win');
+
+// start button and arrow keys events
+button.addEventListener('click', () => {
+  score = 0;
+  document.querySelector('.game-score').innerText = score;
+
+  if (button.classList.contains('start')) {
+    button.classList.remove('start');
+    button.classList.add('restart');
+    button.innerText = 'Restart';
+    startMessage.classList.add('hidden');
+    setGame();
+  } else {
+    loseMessage.classList.add('hidden');
+    setGame();
+  }
+});
+
+document.addEventListener('keyup', (e) => {
+  switch (e.code) {
+    case 'ArrowLeft':
+      slideLeft();
+      setNewNumber();
+      break;
+
+    case 'ArrowRight':
+      slideRight();
+      setNewNumber();
+      break;
+
+    case 'ArrowUp':
+      slideUp();
+      setNewNumber();
+      break;
+
+    case 'ArrowDown':
+      slideDown();
+      setNewNumber();
+      break;
+  }
+
+  document.querySelector('.game-score').innerText = score;
+});
+
+// setting new game
+function setGame() {
+  gameField = [
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+  ];
+
+  for (let r = 0; r < rows; r++) {
+    for (let c = 0; c < columns; c++) {
+      const num = gameField[r][c];
+
+      updateCell(field.rows[r].cells[c], num);
+    }
+  }
+
+  setNewNumber();
+  setNewNumber();
+}
+
+// setting new 2 or 4 in random empty cells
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,258 @@
...
+    for (let r = 0; r < rows; r++) {
+      gameField[r][c] = row[r];
+
+      const num = gameField[r][c];
+
+      updateCell(field.rows[r].cells[c], num);
+    }
+  }
+}
+
```
Virass, line 239: Use DRY principle. Think about how you can merge these functions into one.

src/scripts/main.js
```diff
@@ -1,3 +1,258 @@
...
+    row.reverse();
+    gameField[r] = row;
+
+    for (let c = 0; c < columns; c++) {
+      const num = gameField[r][c];
+
+      updateCell(field.rows[r].cells[c], num);
+    }
+  }
+}
```
Virass, line 194: Same this. Doubling code

Virass (changes requested):
Cool job, Great readability code, but fix several moment (doubling code) and change mechanics. When i shifting block to any side, but blocks not moved, script not must be create new block

image

src/scripts/main.js
rafmamedov: Thank you for your constructive review. Although it took me some time to think about and find the solution, it allowed me to make my code shorter and clearer!

rafmamedov:
vsdv

src/scripts/main.js
rafmamedov: Got it!

src/scripts/main.js
```diff
@@ -1,3 +1,225 @@
...
+
+    for (let r = 0; r < rows; r++) {
+      gameField[r][c] = row[r];
+
+      const num = gameField[r][c];
+
+      updateCell(field.rows[r].cells[c], num);
+    }
+  }
+}
```
DarkMistyRoom, line 205: It is better, but you still repeat the code in these two functions. think about the `transpose` function and merging this logic

DarkMistyRoom (changes requested):

src/scripts/main.js
rafmamedov: Thank you!

src/scripts/main.js
```diff
@@ -1,3 +1,212 @@
...
+  const currentField = [...gameField].toString();
+
+  if (previousField !== currentField) {
+    setNewNumber();
+  }
+
+  document.querySelector('.game-score').innerText = score;
+});
+
+// setting new game
```
loralevitska, line 54: Remove comments. Fix all cases please

src/scripts/main.js
```diff
@@ -1,3 +1,212 @@
...
+  const emptyCell = gameField.some(r => r.some(c => c === 0));
+
+  if (emptyCell) {
+    return;
+  }
+
+  for (let r = 0; r < rows; r++) {
+    for (let c = 0; c < columns - 1; c++) {
+      if (gameField[r][c] === gameField[r][c + 1]
+        || gameField[c][r] === gameField[c + 1][r]) {
```
loralevitska, line 205: Long expressions in `if` move to variables with meaningful names. Fix all cases

src/scripts/main.js
```diff
@@ -1,3 +1,212 @@
...
+  }
+}
+
+function transpose() {
+  gameField = gameField[0].map((_, i) => gameField.map(newRow => newRow[i]));
+}
+
+// game over message
+function setGameOver() {
+  const emptyCell = gameField.some(r => r.some(c => c === 0));
```
loralevitska, line 196:
```suggestion
  const emptyCell = gameField.some(r => r.some(c => !c));
```

src/scripts/main.js
```diff
@@ -1,3 +1,212 @@
...
+      row.reverse();
+    }
+
+    gameField[r] = row;
+  }
+}
+
+function updateField() {
+  for (let r = 0; r < rows; r++) {
+    for (let c = 0; c < columns; c++) {
```
loralevitska, line 182: Try to use more meaningful naming then everything will be clear when reading the code without unnecessary comments. Fix all cases

```suggestion
  for (let row = 0; r < rows; r++) {
    for (let cell = 0; c < columns; c++) {
```

loralevitska (changes requested):
Please deploy your solution, and before review check demo_link
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review
image

nazarmatsevych (approved):
🔥

PR #175, opened 2023-01-13
first try

RudBeee:
https://RudBeee.github.io/js_2048_game


nazarmatsevych (approved):
Great work,
just small suggestion:
restart button should restart game, not to stop it

image

PR #179, opened 2023-01-15
solution

AnnyKoval:
- [DEMO LINK](https://AnnyKoval.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,243 @@
 'use strict';
 
-// write your code here
+let score = 0;
+let isPlaying = false;
+
+const row1 = [0, 1, 2, 3];
+const row2 = [4, 5, 6, 7];
+const row3 = [8, 9, 10, 11];
+const row4 = [12, 13, 14, 15];
+
+const column1 = [0, 4, 8, 12];
+const column2 = [1, 5, 9, 13];
+const column3 = [2, 6, 10, 14];
+const column4 = [3, 7, 11, 15];
+
+const rightDirection = [column3, column2, column1].flat();
+const leftDirection = [column2, column3, column4].flat();
+const upDirection = [row2, row3, row4].flat();
+const downDirection = [row3, row2, row1].flat();
+
+const cells = document.querySelectorAll('.field-cell');
+const startButton = document.querySelector('.start');
+const restartButton = document.createElement('button');
+const messageLoseElement = document.querySelector('.message-lose');
+const messageWinElement = document.querySelector('.message-win');
+const messageStartElement = document.querySelector('.message-start');
+
+restartButton.classList.add('button', 'restart');
+restartButton.textContent = 'Restart';
+restartButton.hidden = true;
+document.querySelector('.controls').append(restartButton);
+
+const checkIfHaveMoves = () => {
+  if ([...cells].filter(cell => cell.classList.length === 1).length > 0) {
+    return true;
+  }
+
+  let haveMoves = false;
+
+  cells.forEach((cell, i) => {
+    const topCell = row1.includes(i) ? null : cells[i - 4];
+    const bottomCell = row4.includes(i) ? null : cells[i + 4];
+    const leftCell = column1.includes(i) ? null : cells[i - 1];
+    const rightCell = column4.includes(i) ? null : cells[i + 1];
+
+    const adjacentCells = [topCell, bottomCell, leftCell, rightCell];
+
+    adjacentCells.forEach(adjacentCell => {
+      if (!adjacentCell) {
+        return;
+      }
+
+      if (adjacentCell.textContent === cell.textContent) {
+        haveMoves = true;
+      }
+    });
+  });
+
+  return haveMoves;
+};
+
+const updateScore = (value) => {
+  if (value) {
+    score += value;
+  } else {
+    score = 0;
+  }
+  document.querySelector('.game-score').textContent = score;
+};
+
+const findCellForNumber = function() {
+  let found = false;
+  let cellIndex;
+
+  while (!found) {
+    cellIndex = Math.floor(Math.random() * 16);
+
+    if (cells[cellIndex].classList.length === 1) {
+      found = true;
+    }
+  }
+
+  return cells[cellIndex];
+};
+
+const generateNumberForTile = () => {
+  const num = Math.random();
+
+  return num < 0.9 ? 2 : 4;
+};
+
+const appendNewNumberTile = () => {
+  const number = generateNumberForTile();
+  const cell = findCellForNumber();
+
+  cell.classList.add(`field-cell--${number}`);
+  cell.textContent = number;
+};
+
+const updateCell = (cell, value = '') => {
+  cell.className = value ? `field-cell field-cell--${value}` : 'field-cell';
+  cell.textContent = value;
+};
+
+const moveHelper = (currentCell, nextCell, cellValue) => {
+  if (nextCell.classList.length > 1) {
+    return;
+  };
+
+  updateCell(currentCell);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+  messageWinElement.classList.add('hidden');
+
+  for (const cell of cells) {
+    cell.className = 'field-cell';
+    cell.textContent = '';
+  };
+  innit();
+};
+
+const documentCallbak = (e) => {
```
witflash, line 198: `keydownCallback` if you use that names

src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+  if (!checkIfHaveMoves()) {
+    isPlaying = false;
+    messageLoseElement.classList.remove('hidden');
+  };
+
+  const currentCellsValues = [...cells].map(cell => cell.textContent);
+
+  if (!isPlaying) {
+    return;
+  }
```
witflash, line 210: Why do you need this check? You can do `return` inside `if` block upper.

src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+
+const documentCallbak = (e) => {
+  e.preventDefault();
+
+  if (!checkIfHaveMoves()) {
+    isPlaying = false;
+    messageLoseElement.classList.remove('hidden');
+  };
+
+  const currentCellsValues = [...cells].map(cell => cell.textContent);
```
witflash, line 206: You don't need to create a new array here (`[...cells]`) because `.map` method is not mutable and returns a new array by itself.

src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+    case 'ArrowRight':
+      moveDirectionAndSum(rightDirection, 'right');
+      break;
+  };
+
+  if (!isPlaying) {
+    return;
+  }
+
+  const newCellsValues = [...cells].map(cell => cell.textContent);
```
witflash, line 231: And here you can use just `cells.map(cell => ...`

src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+const messageWinElement = document.querySelector('.message-win');
+const messageStartElement = document.querySelector('.message-start');
+
+restartButton.classList.add('button', 'restart');
+restartButton.textContent = 'Restart';
+restartButton.hidden = true;
+document.querySelector('.controls').append(restartButton);
+
+const checkIfHaveMoves = () => {
+  if ([...cells].filter(cell => cell.classList.length === 1).length > 0) {
```
witflash, line 34: Use just `cells.filter` because `.filter` method does not mutate an array.

witflash (changes requested):
Good job!
Check my comments, please

src/scripts/main.js
AnnyKoval: Added requested changes

src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+      moveDirectionAndSum(leftDirection, 'left');
+      break;
+    case 'ArrowRight':
+      moveDirectionAndSum(rightDirection, 'right');
+      break;
+  };
+
+  if (!isPlaying) {
+    return;
+  }
```
IvanFesenko, line 229: Previous comment about this check, why do you need it? isPlaying can't be changed on lines 210 - 225

IvanFesenko (changes requested):
Looks like yor game didn't work(

src/scripts/main.js
AnnyKoval: why it doesn't work if I made changes as of requested?

AnnyKoval: it fixed now with [...], check-up, please

src/scripts/main.js
OlEzhA0: not fixed yet
```suggestion
  if (cells.filter(cell => cell.classList.length === 1).length > 0) {
```
or even
```suggestion
  if (cells.some(cell => cell.classList.length === 1)) {
```

Because why do we need to do iterations if we found at least 1 element

src/scripts/main.js
```diff
@@ -1,3 +1,239 @@
...
+
+  return cells[cellIndex];
+};
+
+const generateNumberForTile = () => {
+  const num = Math.random();
+
+  return num < 0.9 ? 2 : 4;
+};
```
OlEzhA0, line 90: in mobile 2048, `8` digit also can be generated
you can do something like

```
const numbers = [2, 4, 8]
const randomIdx = Math.floor(Math.random() * 3)

return numbers[randomIdx]
```

src/scripts/main.js
```diff
@@ -1,3 +1,239 @@
...
+
+  while (!found) {
+    cellIndex = Math.floor(Math.random() * 16);
+
+    if (cells[cellIndex].classList.length === 1) {
+      found = true;
+    }
+  }
+
+  return cells[cellIndex];
```
OlEzhA0, line 83: you don't handle case when you haven't any empty cell
try to collect all empty cells and choose some empty with `Math.random()`
```suggestion
  while (true) {
    const cellIndex = Math.floor(Math.random() * 16);

    if (cells[cellIndex].classList.length === 1) {
      return cells[cellIndex];
    }
  }
```

src/scripts/main.js
```diff
@@ -1,3 +1,239 @@
...
+  if (nextCell.classList.length > 1) {
+    return;
+  };
+
+  updateCell(currentCell);
+  updateCell(nextCell, cellValue);
+};
+
+const addHelper = (currentCell, nextCell, cellValue) => {
+  if (Number(nextCell.textContent) === Number(cellValue)) {
```
OlEzhA0, line 115: more preferred syntax 
```suggestion
  if (+nextCell.textContent === +cellValue) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,239 @@
...
+    updateCell(currentCell);
+    updateCell(nextCell, cellValue * 2);
+    updateScore(cellValue * 2);
+
+    if (Number(cellValue * 2) === 2048) {
```
OlEzhA0, line 120: move `+cellValue * 2` to constant

src/scripts/main.js
```diff
@@ -1,3 +1,239 @@
...
+    case 'left':
+      step = -1;
+      break;
+    case 'up':
+      step = -4;
+      break;
+    case 'down':
+      step = 4;
+      break;
+  };
```
OlEzhA0, line 143: some trick for you
```suggestion
  const config = {
    right: 1,
    left: -1,
    up: -4,
    down: 4,
  }

   const step = config[direction];
```

src/scripts/main.js
```diff
@@ -1,3 +1,239 @@
...
+  });
+};
+
+const moveDirectionAndSum = (directionArr, direction) => {
+  moveAddTiles(directionArr, direction);
+  moveAddTiles(directionArr, direction);
+  moveAddTiles(directionArr, direction);
+  moveAddTiles(directionArr, direction, 'add');
+  moveAddTiles(directionArr, direction);
+  moveAddTiles(directionArr, direction);
```
OlEzhA0, line 170: why do we need to call this function exactly this times?

src/scripts/main.js
```diff
@@ -1,3 +1,239 @@
...
+};
+
+const innit = () => {
```
OlEzhA0, line 173:
```suggestion
const init = () => {
```

OlEzhA0 (changes requested):
hey! please check all unresolved comments

IvanFesenko (approved):
Goof job

PR #183, opened 2023-01-20
Solution

OksanaBaloh:
[DEMO LINK](https://OksanaBaloh.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -21,11 +21,7 @@ for (const row of rows) {
   }
 }
 
-let cells = [];
-
-for (const row of rows) {
-  cells = cells.concat(row);
-}
+const cells = rows.flat();
 
 startBtn.addEventListener('click', () => {
   startMsg.classList.add('hidden');


src/scripts/main.js
```diff
@@ -1,3 +1,206 @@
...
+// Variables
 
-// write your code here
+const startBtn = document.querySelector('.start');
+const score = document.querySelector('.game-score');
+const startMsg = document.querySelector('.message-start');
+const winMsg = document.querySelector('.message-win');
+const loseMsg = document.querySelector('.message-lose');
+
+// Field Variables
```
loralevitska, line 10: Remove comment. Try it write obvious code that explain all we need without comments)
Fix all cases

src/scripts/main.js
```diff
@@ -1,3 +1,206 @@
...
+        break;
+      case 'ArrowRight':
+        move(rows, rightOrDown);
+        break;
+      case 'ArrowUp':
+        move(columns, leftOrUp);
+        break;
+      case 'ArrowDown':
+        move(columns, rightOrDown);
+        break;
```
loralevitska, line 74: switch case better use with `default`

src/scripts/main.js
```diff
@@ -1,3 +1,206 @@
...
+    }
+  });
+});
+
+// Functions
+
+function addNumber(allCells) {
+  const emptyCells = allCells.filter(cell => cell.textContent === '');
```
loralevitska, line 82:
```suggestion
  const emptyCells = allCells.filter(cell => !cell.textContent);
```

src/scripts/main.js
```diff
@@ -1,3 +1,206 @@
...
+
+  if (emptyCells.length === 0) {
```
loralevitska, line 84:
```suggestion
  if (!emptyCells.length) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,206 @@
...
+    emptyCells[index].textContent = '4';
+  }
+}
+
+function styleCells(allCells) {
+  for (const cell of allCells) {
+    cell.className = '';
+    cell.classList.add('field-cell');
+
+    if (cell.textContent !== '') {
```
loralevitska, line 103:
```suggestion
    if (cell.textContent) {
```

loralevitska (changes requested):

1. Remove default outline 
image
2. I guess better remove congrats text if user lose or restart game
image
image

src/scripts/main.js
```diff
@@ -1,3 +1,203 @@
...
+    columns[i].push(row[i]);
+  }
+}
+
+let cells = [];
+
+for (const row of rows) {
+  cells = cells.concat(row);
+}
+
```
Volodymyr-Melnyk-Fron10der, line 29: There is method like `arr.flat()` 

```
const arr1 = [0, 1, 2, [3, 4]];

console.log(arr1.flat());
// Expected output: Array [0, 1, 2, 3, 4]
```

You can use it as well `const cells = rows.flat()`

Volodymyr-Melnyk-Fron10der (changes requested):
Excellent work! But I see one more improvement )

src/scripts/main.js
OksanaBaloh: Nice method, thank you)

ViktorHavryliuk (approved):

PR #185, opened 2023-01-23
Develop

diolna:
- [DEMO LINK](https://diolna.github.io/js_2048_game/)


vlad-demchuk (changes requested):
I can't lose. I like it, but you should fix it)
Jan-24-2023 09-45-26

diolna:
Added button lock to work properly when pressed very quickly. Disabled the delay in the appearance of a new number in order to pass tests for scoring

PR #189, opened 2023-01-27
Develop

Mykola-Busarhin:
[DEMO LINK](https://Mykola-Busarhin.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,250 @@
 'use strict';
 
-// write your code here
+const gameSize = 4;
+const startBtn = document.querySelector('.button');
+const startMessage = document.querySelector('.message-start');
+const loseMessage = document.querySelector('.message-lose');
+const scoreField = document.querySelector('.game-score');
+let score = 0;
+let gameSquare = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+
+];
+
+startBtn.addEventListener('click', () => {
+  newGame();
+
+  // const btn = e.target;
+  // console.log(startBtn.innerText);
+
+  if (startBtn.innerText === 'Start') {
+    // console.log('true');
+    startBtn.innerText = 'Restart';
+    startBtn.classList.replace('start', 'restart');
+  } else if (startBtn.textContent === 'Restart') {
+    startBtn.innerText = 'Start';
+    startBtn.classList.replace('restart', 'start');
+  }
+
+  // startBtn.classList.toggle('')
+
+  // if (startBtn.textContent === 'Restart') {
+  // startBtn.innerText = 'Start';
+  // startBtn.classList.replace('restart', 'start');
+  // }
+
+  // btn.innerText = 'Restart';
+  // btn.classList.replace('start', 'restart');
+  startMessage.classList.add('hidden');
+});
+
+function generateNewNumber() {
+  let randomRow;
+  let randomCell;
+
+  do {
+    randomRow = Math.floor(Math.random() * gameSize);
+    randomCell = Math.floor(Math.random() * gameSize);
+
+    if (gameSquare[randomRow][randomCell] === 0) {
+      gameSquare[randomRow][randomCell] = Math.random() > 0.1 ? 2 : 4;
+      break;
+    }
+  } while (true);
+}
+
+function fillGameField() {
+  const gameBoard = document.querySelector('.game-field');
+  const rows = [...gameBoard.rows];
+  const e = [];
+
+  for (const a of rows) {
+    e.push([...a.cells]);
+  }
+
+  for (let i = 0; i < gameSize; i++) {
+    for (let x = 0; x < gameSize; x++) {
+      const gameValue = e[i][x];
+      const arrayValue = gameSquare[i][x];
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,224 @@
...
+    gameSquare[randomRow][randomCell] = Math.random() > 0.1 ? 2 : 4;
+  } else {
+    generateNewNumber();
+  }
+}
+
+function fillGameField() {
+  const gameBoard = document.querySelector('.game-field');
+  const rows = [...gameBoard.rows];
+  const e = [];
```
loralevitska, line 45: What is `e`? Try to use more meaningful variable names

src/scripts/main.js
```diff
@@ -1,3 +1,224 @@
...
+
+  for (const a of rows) {
+    e.push([...a.cells]);
+  }
+
+  for (let i = 0; i < gameSize; i++) {
+    for (let x = 0; x < gameSize; x++) {
+      const gameValue = e[i][x];
```
loralevitska, line 53: Same here. Nothing is clear, because instead of variables there are only letters

src/scripts/main.js
```diff
@@ -1,3 +1,224 @@
...
+      const arrayValue = gameSquare[i][x];
+
+      gameValue.innerText = arrayValue === 0 ? '' : arrayValue;
```
loralevitska, line 56: If you need to check just for the value, then you can write it down simpler.... If you need only for zero, and an empty string or `undefined` can pass, then leave the solution that was

```suggestion
      gameValue.innerText = !arrayValue ? '' : arrayValue;
```

src/scripts/main.js
```diff
@@ -1,3 +1,224 @@
...
+      break;
+    case 'ArrowUp':
+      moveRowsUp();
+      generateNewNumber();
+      break;
+    case 'ArrowDown':
+      moveRowsDown();
+      generateNewNumber();
+      break;
+  }
```
loralevitska, line 99:
```suggestion
  switch (pressedKey) {
    case 'ArrowRight':
      moveRowsRight();
      generateNewNumber();
      break;
      
    case 'ArrowLeft':
      moveRowsLeft();
      generateNewNumber();
      break;
      
    case 'ArrowUp':
      moveRowsUp();
      generateNewNumber();
      break;
      
    case 'ArrowDown':
      moveRowsDown();
      generateNewNumber();
      break;
  }
```

src/scripts/main.js
```diff
@@ -1,3 +1,224 @@
...
+    newRow.reverse();
+
+    gameSquare[0][i] = newRow[0];
+    gameSquare[1][i] = newRow[1];
+    gameSquare[2][i] = newRow[2];
+    gameSquare[3][i] = newRow[3];
+  };
+
+  return gameSquare;
+}
```
loralevitska, line 191: These two functions look very similar, merge the repeating code into a helper function

loralevitska (changes requested):
1. Remove default outline 
image

src/scripts/main.js
```diff
@@ -1,3 +1,221 @@
...
+  for (let row = 0; row < gameSize; row++) {
+    for (let column = 0; column < gameSize; column++) {
+      const gameValue = gameBoardCells[row][column];
+      const arrayValue = gameSquare[row][column];
+
+      gameValue.innerText = arrayValue === 0 ? '' : arrayValue;
+      gameValue.classList.value = '';
+
+      gameValue.classList.add('field-cell',
+        `field-cell--${gameValue.innerText}`);
```
DarkMistyRoom, line 61:
```suggestion
      gameValue.classList.add(
        'field-cell',
        `field-cell--${gameValue.innerText}`
      );
```

src/scripts/main.js
```diff
@@ -1,3 +1,221 @@
...
+  }
+}
+
+function newGame() {
+  gameSquare = [
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+  ];
```
DarkMistyRoom, line 72: use a variable with default state

src/scripts/main.js
```diff
@@ -1,3 +1,221 @@
...
+  fillGameField();
+}
+
+document.addEventListener('keydown', (e) => {
+  const pressedKey = e.code;
+
+  switch (pressedKey) {
+    case 'ArrowRight':
+      moveRowsRight();
+      generateNewNumber();
```
DarkMistyRoom, line 86: could generateNewNumber outside the switch to avoid duplicating?

DarkMistyRoom (changes requested):
Good, but this require isn't suttisfied. 
image

there was this situation on the field except for the highlighted cell. I clicked the bottom arrow and it appeared

and there should appear 2 cell in the begining

Mykola-Busarhin:
Hi, didn`t quite get your comment about veriable with default state for lines 67-72.

src/scripts/main.js
SanyaBratashchuk: move `[
    [0, 0, 0, 0],
    [0, 0, 0, 0],
    [0, 0, 0, 0],
    [0, 0, 0, 0],
  ] ` this part to a variable and use it when you need

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #190, opened 2023-01-27
need help

Ivanenko1402:
[DEMO LINK](https://Ivanenko1402.github.io/js_2048_game/)


agniya-i (changes requested):
In case game over message should be shown if there are no more available moves, there is just empty board now

Screenshot 2023-01-28 at 13 04 12

PR #194, opened 2023-01-30
first attempt

metinbicaksiz:
[DEMO LINK](https://metinbicaksiz.github.io/js_2048_game/)


polosanya (changes requested):
1. I still can play after loss, and restart button doesn't work
2048-bug
2. Restart doesn't work at all
2048-restart

PR #195, opened 2023-01-31
game 2048

webdevw:
- [DEMO LINK](https://webdevw.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,275 @@
 'use strict';
 
-// write your code here
+const tableGame = document.querySelector('.game-field');
+const rows = tableGame.rows;
+const button = document.querySelector('button');
+const score = document.querySelector('.game-score');
+
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const cells = document.querySelectorAll('.field-cell');
+
+let table2 = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
+
+button.style.outline = 'none';
+
+let starting = false;
+
+button.addEventListener('click', action => {
+  function toggleBtn(text = '') {
+    button.textContent = text;
+    button.classList.toggle('start');
+    button.classList.toggle('restart');
+  }
+
+  messageStart.classList.add('hidden');
+  messageWin.classList.toggle('hidden', true);
+  messageLose.classList.toggle('hidden', true);
+
+  function startGame() {
+    starting = true;
+    score.textContent = 0;
+
+    table2 = [
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+    ];
+
+    addInRandomCell(table2);
+    addInRandomCell(table2);
+    addValuesToCells(table2);
+    changeStyleCells(cells);
+  }
+
+  if (button.classList.contains('start')) {
+    toggleBtn('Restart');
+    startGame();
+  } else {
+    startGame();
+  }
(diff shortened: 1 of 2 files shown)


qoa11a (changes requested):
here's a bug
when I make right/left move two 16s should collapse and become 32. But you can see what is happening in the video

https://user-images.githubusercontent.com/102293862/216020731-a9a9afb8-699a-40bc-80c2-d25a8d22f2f3.mov

src/scripts/main.js
```diff
@@ -1,3 +1,286 @@
...
+  const emptyCellsIndexes = [];
+  const randomNumber = Math.round(Math.random() * 100);
+  const value = randomNumber > percent
+    ? 2
+    : 4;
+
+  for (let i = 0; i < rows.length; i++) {
+    for (let j = 0; j < rows.length; j++) {
+      if (matrix[i][j] === 0) {
+        emptyCellsIndexes.push(`${i}${j}`);
```
slavanosatskyi, line 234:
```suggestion
        emptyCellsIndexes.push([i, j]);
```
as an option you could do it like this

src/scripts/main.js
```diff
@@ -1,3 +1,286 @@
...
+const column = 4;
+
+let table2 = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
+
+button.style.outline = 'none';
```
slavanosatskyi, line 21: this has to be moved to css

src/scripts/main.js
```diff
@@ -1,3 +1,286 @@
...
+const tableGame = document.querySelector('.game-field');
+const rows = tableGame.rows;
+const button = document.querySelector('button');
+const score = document.querySelector('.game-score');
+
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const cells = document.querySelectorAll('.field-cell');
+const column = 4;
```
slavanosatskyi, line 12: better to name is such a way that it contains more precise meaning like columnsCount or columnsNumber. Also most likely that in this app is should always equal to tableGame.rows.length

src/scripts/main.js
```diff
@@ -1,3 +1,286 @@
...
+    clObj[i] = obj[i];
+  }
+
+  return clObj;
+}
+
+function checkOnWin(matrix) {
+  for (const row of matrix) {
+    for (const cell of row) {
+      if (cell === 2048) {
```
slavanosatskyi, line 267: 2048 can be moved to constants with a meaningful name something like victoryValue, finalValue, etc..

slavanosatskyi (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,285 @@
...
+const button = document.querySelector('button');
+const score = document.querySelector('.game-score');
+
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const cells = document.querySelectorAll('.field-cell');
+const winValue = 2048;
+
+let table2 = [
```
Anastasiia-Svintsova, line 15: Why is it table2? 
```suggestion
let table = [
```

src/scripts/main.js
```diff
@@ -1,3 +1,285 @@
...
+    return newRow;
+  }
+
+  const copy1 = [];
+
+  for (let j = 0; j < columnLength; j++) {
+    let row = copy[j];
+
+    row = slide(row);
+    copy1[j] = row;
```
Anastasiia-Svintsova, line 200: Try not to use numbers in the variable names. Better to use the words `firstCopy`, `secondCopy` for example

Anastasiia-Svintsova (approved):

PR #202, opened 2023-02-07
finished

barantarasNEW:
https://barantarasNEW.github.io/js_2048_game/

Не знаю як ще вкоротити код... Якщо вкорочувати ті напрямки, мені здається буде не читабельним.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,331 @@
 'use strict';
 
-// write your code here
+const game = {
+  matrix: [
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+  ],
+  score: 0,
+  moved: false,
+};
+
+const gameFieldBody = document.querySelector('.game-field tbody');
+const controls = document.querySelector('.controls');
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,331 @@
...
+  render();
+
+  if (game.moved) {
+    setTimeout(() => {
+      gameScore.textContent = game.score;
+
+      addRandomCell();
+
+      render();
+    }, 240);
```
DarkMistyRoom, line 315: I think it isn't a good idea to use delay here because I can make more than one move before the cell will appear

src/scripts/main.js
```diff
@@ -1,3 +1,331 @@
...
+        let currRow = row;
+
+        while (currRow < 3 && game.matrix[col][currRow + 1] === 0) {
+          currRow++;
+        }
+
+        checkEmptyTail(col, row, col, currRow);
+      }
+    }
+  }
```
DarkMistyRoom, line 158: too much duplication here. think about moving this logic to a function. also, you can combine this logic with `checkSameTail`

DarkMistyRoom (changes requested):

IvanFesenko (approved):
Well done.

PR #216, opened 2023-02-28
2048

Igor-Yaremchuk:
[2048](https://Igor-Yaremchuk.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,242 @@
 'use strict';
 
-// write your code here
+const fieldRows = document.querySelector('tbody').rows;
+const scoreBoard = document.querySelector('.game-score');
+const startBtn = document.querySelector('.start');
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,242 @@
...
+  messageWin.classList.add('hidden');
+  messageLose.classList.add('hidden');
+
+  startBtn.classList.add('restart');
+  startBtn.innerText = 'Restart';
+});
+
+document.addEventListener('keyup', type => {
+  if (startBtn.classList.contains('restart')) {
+    if (type.code === 'ArrowLeft') {
```
polosanya, line 46: You can use switch...case instead as you have limited amount of type.codes

polosanya (changes requested):
Almost done! Let's improve your code a bit)

anstsot (approved):
Great job, approved!

PR #221, opened 2023-03-07
2048

Tanya-Grebenkina:
[DEMO LINK](https://Tanya-Grebenkina.github.io/js_2048_game/)


OlEzhA0 (changes requested):
Hey! check please rules of original 2048 game
in your implementation I can do some move even when I haven’t it

in this video I pressed only `right arrow` 

https://user-images.githubusercontent.com/60295799/223662406-d8ab310b-cf12-4a65-a944-dd4967e43569.mov

Tanya-Grebenkina:
Oh, man, well, I'm going to go deal with all of this.🦦

tetiana-ivanchenko (changes requested):
this problem still didn't fixed
image

src/styles/main.scss
```diff
@@ -15,11 +15,12 @@ body {
...
   height: 75px;
   border-radius: 5px;
-  color: #776e65;
-  box-sizing: border-box;
   text-align: center;
   vertical-align: center;
+  color: #776e65;
+  box-sizing: border-box;
   user-select: none;
+  transition: all 0.3s ease;
```
SerhiiKirik, line 23: don't use it like that, because it's the same as transition-duration 0.3s
It's too hard for the browser
Just list all params that will be chenged

SerhiiKirik (approved):
LGTM 🔥

PR #223, opened 2023-03-08
Feat: Implement 2048 game

mykyta-krainik:
[DEMO LINK](https://mykyta-krainik.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,3 +1,279 @@
...
+  return newRow;
+}
+
+function slideUp() {
+  for (let c = 0; c < columnsNum; c++) {
+    const row = slide([
+      field[0][c],
+      field[1][c],
+      field[2][c],
+      field[3][c],
```
agniya-i, line 63: one small thing that I noticed, it's not a good idea to hardcode these indexes, what if we have a field with 5x5? 🙂

agniya-i (approved):
Good job 🔥

PR #229, opened 2023-03-13
Solution

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/js_2048_game/)


qoa11a (changes requested):
Good job! 👍 
One small fix:
You should reset the score counter after the game's lost and restarted.

witflash (changes requested):
Well done!
But here is one point.
Could you check your logic for such a case?
image
 
If I have a legal movement but press the down key, I've lost the game. 
You have to check if there are any legal movements and don't end the game.

slavanosatskyi (approved):

PR #232, opened 2023-03-15
solution

vladyslav-krainyk:
[DEMO LINK](https://vladyslav-krainyk.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,331 @@
 'use strict';
 
-// write your code here
+const controls = document.querySelector('.controls');
+const startButton = document.querySelector('.start');
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const gameScore = document.querySelector('.game-score');
+const cells = document.querySelectorAll('.field-cell');
+const restartButton = document.createElement('button');
+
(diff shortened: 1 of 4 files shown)


SerhiiKirik (approved):
LGTM 🔥

DarkMistyRoom (changes requested):
- in [this video](https://www.loom.com/share/7d21737e22e24b33bafd0f3957e9832a), two 2 cells should merge but it didn't happen

- in[ this video](https://www.loom.com/share/22731d0b216a4e5e9e2f7a1209ba2c0f) you can see that the bottom 2 cell should go to the 2nd from the top place but it stay on the previous position

image

- also please check the merge logic. I noticed that it doesn't work in some cases

- restart button should start a new game. user doesn't have to click start after that
- why do I lose if there are so many variants for merging?
image

src/scripts/main.js
```diff
@@ -1,3 +1,269 @@
...
+const ROW_SIZE = 4;
+const COL_SIZE = 4;
+
+startButton.addEventListener('click', () => {
+  board = [
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+  ];
```
DarkMistyRoom, line 35: better use constant with default state

DarkMistyRoom (approved):
Good job!

PR #235, opened 2023-03-16
solution

MSA1991:
- [DEMO LINK](https://msa1991.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,281 @@
 'use strict';
 
-// write your code here
+const buttonAction = document.querySelector('.button');
+const messages = document.querySelectorAll('.message');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+const gameScore = document.querySelector('.game-score');
+const fieldLength = 4;
+
+let score;
+let field;
+
+buttonAction.addEventListener('click', () => {
+  score = 0;
+
+  field = [
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+  ];
+
+  addNewNumber();
+  addNewNumber();
+
+  AllMessagesHidden();
+
+  buttonAction.innerHTML = 'Reset';
+  buttonAction.classList.remove('start');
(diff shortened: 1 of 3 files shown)


loralevitska (changes requested):
Almost perfect) Let's improve
- This message should be visible when 2048 value is displayed in any cell (not score)
image

src/scripts/main.js
```diff
@@ -1,3 +1,294 @@
...
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+  ];
+
+  addNewNumber();
+  addNewNumber();
+
+  allMessagesHidden();
```
DarkMistyRoom, line 31:
```suggestion
  hideAllMessages();
```

DarkMistyRoom: start functions names with a verb

src/scripts/main.js
```diff
@@ -1,3 +1,294 @@
...
+
+    row = filterZero(row);
+
+    addZeros(row);
+
+    field[i] = row;
+  }
+
+  addNewNumber();
+  render();
```
DarkMistyRoom, line 139: you can move these actions to `keyup `handler to avoid duplication
image

DarkMistyRoom (changes requested):
- new cell should appear only after moving. this crossed cell appeared after pressing the move button with no changes

image

- and sometimes new cell doesn't appear after moving(or appears in an already occupied place). try to reproduce it and fix it. Check the logic of `addNewNumber` function

src/scripts/main.js
```diff
@@ -1,3 +1,288 @@
...
+const buttonAction = document.querySelector('.button');
+const messages = document.querySelectorAll('.message');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+const gameScore = document.querySelector('.game-score');
+const fieldLength = 4;
+const up = 'ArrowUp';
+const down = 'ArrowDown';
+const left = 'ArrowLeft';
+const right = 'ArrowRight';
```
OleziO, line 12: Combine to one object.

OleziO (approved):

PR #237, opened 2023-03-17
Ready

sergik94:
https://sergik94.github.io/js_2048_game/


reports/report.html
loralevitska, line 1: For future don't add such files for review. Commit only solution

src/scripts/main.js
```diff
@@ -1,3 +1,375 @@
...
+
+  activeCell.innerHTML = currCell.innerText;
+  activeCell.dataset.currNumb = currCell.dataset.currNumb;
+  activeCell.style.transitionDuration = cellTransDur + 'ms';
+  supportField.append(activeCell);
+  currCell.className = 'field-cell';
+  currCell.innerHTML = '';
+  currCell.dataset.currNumb = '';
+
+  if (e.keyCode === 37 || e.keyCode === 39) {
```
loralevitska, line 115: Move this keyCodes to constants

src/scripts/main.js
```diff
@@ -1,3 +1,375 @@
...
+    activeCell.style.left = newCell.offsetLeft + 'px';
+  }
+
+  if (e.keyCode === 38 || e.keyCode === 40) {
```
loralevitska, line 119: Same here

src/scripts/main.js
```diff
@@ -1,3 +1,375 @@
...
+    activeCell.style.top = newCell.offsetTop + 'px';
+  }
+
+  if (newCell.dataset.currNumb === activeCell.dataset.currNumb) {
+    newCell.style.transitionDuration = cellTransDur + 'ms';
+    newCell.dataset.currNumb *= 2;
```
loralevitska, line 125: Why 2? Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

src/scripts/main.js
```diff
@@ -1,3 +1,375 @@
...
+            if (prevCell.dataset.currNumb === currCell.dataset.currNumb) {
+              if (!mergeCell) {
+                mergeCell = true;
+              } else {
+                mergeCell = false;
+                k++;
+              }
+              break;
+            }
+          } while (k > 0);
```
loralevitska, line 239: I see some repeatable code. All duplicates move to helpers (function/variable)

loralevitska (approved):
Amazing

PR #238, opened 2023-03-18
all tests good

Mikhaylo-Orlov:
[DEMO LINK](https://Mikhaylo-Orlov.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,242 @@
 'use strict';
 
-// write your code here
+const score = document.querySelector('.game-score');
+const cell = document.querySelectorAll('.field-cell');
+const button = document.querySelector('.button');
+
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const messageStart = document.querySelector('.message-start');
+
+const rows = 4;
+const columns = 4;
+
+button.addEventListener('click', () => {
+  if (button.textContent === 'Start') {
+    button.classList.remove('start');
+    button.classList.add('restart');
+    button.textContent = 'Restart';
+  } else {
+    button.classList.remove('restart');
+    button.classList.add('start');
+    button.textContent = 'Start';
+  }
+
+  score.textContent = '0';
+  messageWin.classList.add('hidden');
+  messageLose.classList.add('hidden');
+  messageStart.classList.add('hidden');
+
+  beginGame();
+});
+
+let board;
+
+function beginGame() {
+  board = [
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+  ];
+
+  let i = 0;
+
+  for (let r = 0; r < rows; r++) {
+    for (let c = 0; c < columns; c++) {
+      cell[i].id = `${r}${c}`;
+
+      const num = board[r][c];
+
+      updateTile(cell[i], num);
+      i++;
+    }
+  }
+
+  setRandom();
+  setRandom();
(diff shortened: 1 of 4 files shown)


TarasHoliuk:
There are still available moves, but `You lose! Restart the game?` message is displaying, and I can't move tiles:

image

src/scripts/main.js
```diff
@@ -1,3 +1,242 @@
...
+
+      const tile = document.getElementById(`${r}${c}`);
+      const number = board[r][c];
+
+      updateTile(tile, number);
+    }
+  }
+};
+
+function loseGame() {
```
TarasHoliuk, line 217:
```suggestion
function checkIsGameLoss() {
```
you're doing a check here

src/scripts/main.js
```diff
@@ -1,3 +1,242 @@
...
+        || board[r][c] === board[r][c + 1]
+      ) {
+        messageLose.classList.remove('hidden');
+        document.removeEventListener('keyup', handler);
+      }
+    }
+  }
+}
+
+function winGame() {
```
TarasHoliuk, line 234:
```suggestion
function checkIsGameWon() {
```

src/scripts/main.js
```diff
@@ -1,3 +1,242 @@
...
+  tile.classList.value = '';
+  tile.classList.add('field-cell');
+
+  if (num > 0) {
+    tile.innerText = num;
+    tile.classList.add(`field-cell--${num}`);
+  }
+}
+
+const handler = (e) => {
```
TarasHoliuk, line 107:
```suggestion
const handler = (event) => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,242 @@
...
+  switch (e.code) {
+    case 'ArrowLeft': {
+      slideLeftRight(true);
+      setTimeout(setRandom, 500);
```
TarasHoliuk, line 111: remove timeout - when the user clicks fast new tiles appears too late (and that might be the reason for the problem with a too-early message about the user loss)

TarasHoliuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,254 @@
...
+
+  setRandom();
+  setRandom();
+  document.addEventListener('keyup', handler);
+}
+
+function hasEmptyTile() {
+  for (let r = 0; r < rows; r++) {
+    for (let c = 0; c < columns; c++) {
+      if (board[r][c] === 0) {
```
OleziO, line 64: Replace with `!board[r][c]`, it will be the same as `board[r][c] === 0`

OleziO: Check other similar cases.

src/scripts/main.js
```diff
@@ -1,3 +1,254 @@
...
+  tile.classList.value = '';
+  tile.classList.add('field-cell');
+
+  if (num > 0) {
+    tile.innerText = num;
+    tile.classList.add(`field-cell--${num}`);
+  }
+}
+
+const handler = (event) => {
```
OleziO, line 107: Try to fix it
image

src/scripts/main.js
```diff
@@ -1,3 +1,254 @@
...
+    }
+
+    case 'ArrowRight': {
+      slideLeftRight(false);
+      setRandom();
+
+      break;
+    }
+
+    case 'ArrowUp': {
```
OleziO, line 123: Move it to a const object with all case values.

OleziO (changes requested):

OleziO (approved):

PR #245, opened 2023-03-27
Solution

DaveBeetle:
- [DEMO LINK](https://DaveBeetle.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,222 @@
 'use strict';
 
-// write your code here
+const field = document.querySelector('tbody');
+const button = document.querySelector('button');
+const scoreField = document.querySelector('.game-score');
+
+const messageStart = document.querySelector('.message-start');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+
+let state = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0]];
+
+let newState;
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,222 @@
...
+
+const messageStart = document.querySelector('.message-start');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+
+let state = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0]];
```
loralevitska, line 15: Fix it everywhere
```suggestion
  [0, 0, 0, 0]
  ];
```

src/scripts/main.js
```diff
@@ -1,3 +1,222 @@
...
+
+let newState;
+let isWin = false;
+let score = 0;
+const fieldMatrix = [...field.children].map(row => [...row.children]);
```
loralevitska, line 20: map method returns new array. It's non-mutating method. You don't need copy
```suggestion
const fieldMatrix = field.children.map(row => [...row.children]);
```

loralevitska (approved):
- Remove outline 
image

src/scripts/main.js
DaveBeetle: Ok)

src/scripts/main.js
DaveBeetle: Ok, thanks)

PR #246, opened 2023-03-28
2048_game

G0odvin:
[DEMO LINK](https://G0odvin.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,242 @@
 'use strict';
 
-// write your code here
+const fieldRows = document.querySelector('tbody').rows;
+const startGame = document.querySelector('.start');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+const messageStart = document.querySelector('.message-start');
+const gameScore = document.querySelector('.game-score');
+
+let score = 0;
+let rows = 4;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,242 @@
...
+  startGame.innerHTML = 'Restart';
+  messageStart.classList.add('hidden');
+  messageLose.classList.add('hidden');
+  messageWin.classList.add('hidden');
+  gameScore.innerHTML = 0;
+
+  gameStart();
+})
+
+document.addEventListener('keyup', (event) => {
```
witflash, line 44: It's better to add listener for `keydown` I think

src/scripts/main.js
```diff
@@ -1,3 +1,242 @@
...
+  if (!startGame.classList.contains('restart')) {
```
witflash, line 45: Add `e.preventDefault()` before you handle the event. Because I have page scroll if I press up/down buttons.

witflash (changes requested):
Good work. Let's make several improvement.

1. You should prevent default behavior of the key pressing.
2. Need to reset score after game restarting.
3. Check the text on Start/Restart button. It's not centered and looks strange.

qoa11a (approved):

PR #254, opened 2023-04-06
Solution

Ssviatt:
- [DEMO LINK](https://Ssviatt.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,319 @@
 'use strict';
 
-// write your code here
+const controls = document.querySelector('.controls');
+const scoreInfo = document.querySelector('.game-score');
+
+const rows = document.querySelectorAll('.field-row');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,319 @@
...
+    matrix.forEach(row => {
+      row.forEach((num, index) => {
+        if (num === 0) {
+          row.splice(index, 1);
+          row.unshift(0);
+        }
+      });
+
+      for (let i = row.length - 1; i >= 0; i--) {
+        if (row[i] > 0 && row[i] === row[i - 1]) {
```
SerhiiKirik, line 43: you repeat row[i] a lot of times
Would be better if you move this to variables
Also, you can use it in all cases

SerhiiKirik (approved):
Well done 🚀

PR #261, opened 2023-04-10
Dev

Ukrainiane-panda:
[DEMO LINK](https://Ukrainiane-panda.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -22,24 +22,25 @@ for (let i = 0; i < arrayCells.length; i++) {
   arrayCells[i] = new Array(tableSize).fill(0);
 }
 
+let handleKeyDown = false;
+
 // Старт або рестарт гри
 button.addEventListener('click', e => {
+  handleKeyDown = true;
+
   if (!button.classList.contains('restart')) {
     button.textContent = 'Restart';
     button.classList.add('restart');
     button.classList.remove('start');
-    // приховую будь-які меседжі
     messageStart.classList.add('hidden');
   } else {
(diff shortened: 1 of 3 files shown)


SerhiiKirik (changes requested):
Great 🔥 
Is something wrong with the algorithm, bc when i have a situation like this - i can't click & the page just freezes
image

src/scripts/main.js
```diff
@@ -1,3 +1,329 @@
...
+
+const tableSize = 4;
+
+const arrayCells = new Array(tableSize);
+
+for (let i = 0; i < arrayCells.length; i++) {
+  arrayCells[i] = new Array(tableSize).fill(0);
+}
+
+let handleKeyDown = false;
```
SerhiiKirik, line 20: Consider changing the variable name `handleKeyDown` to a more descriptive name, like `canHandleKeyDown` since it is a boolean flag 🚩

src/scripts/main.js
```diff
@@ -1,3 +1,329 @@
...
+        if (checkGameOver()) {
+          messageLose.classList.remove('hidden');
+        }
+
+        if (checkWin()) {
+          messageWin.classList.remove('hidden');
+        }
+        break;
+
+      default:
```
SerhiiKirik, line 131: Consider removing the `throw new Error` statement, since the game should not throw an error if a non- arrow key is pressed. Instead, do nothing or log a message to the console. 🎮

SerhiiKirik (approved):
Great 🔥

PR #276, opened 2023-04-20
Develop

AnastassiaH:
[DEMO LINK](https://anastassiah.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,8 +1,11 @@
 'use strict';
+
 const startButton = document.querySelector('.start');
+
 startButton.insertAdjacentHTML('afterend', `
   <button class="button restart hidden">Reset</button>
   `);
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,191 @@
...
+
+startButton.addEventListener('click', () => {
+  fillNewCell();
+  fillNewCell();
+  toggleHidden(startButton, restartButton, messageStart);
+  document.addEventListener('keydown', handleMove);
+});
+
+restartButton.addEventListener('click', () => {
+  [...fieldCells].map(item => {
```
vlad-demchuk, line 49: It's better to use `forEach` instead of `map` here, as you are not using the returned array and just performing a side effect 🤔

vlad-demchuk (approved):

PR #279, opened 2023-04-24
Solution

dianaavem:
[DEMO LINK](https://dianaavem.github.io/js_2048_game/)


Volodymyr-Melnyk-Fron10der (commented):
1. i can move cells even when I lose

loralevitska (approved):
🔥

PR #280, opened 2023-04-27
Added the task solution

oleksusov:
[DEMO LINK](https://oleksusov.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,386 @@
 'use strict';
 
-// write your code here
+const boardSize = 4;
+const gameBoard = document.querySelector('.game-board');
+const gameScore = document.querySelector('.game-score');
+const button = document.querySelector('.button');
+const messageStart = document.querySelector('.message-start');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+
+const handlerSlide = e => {
+  switch (e.code) {
+    case 'ArrowLeft':
+      if (slideLeft()) {
+        setTimeout(function() {
+          createTile();
+        }, 200);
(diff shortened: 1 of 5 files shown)


polosanya (changes requested):
Good job! Let's improve your code a bit)

anstsot (changes requested):
Looks good from a code perspective but your demo link isn't working. could you check it, please?
If you have any questions, feel free to ask them in the chat

src/scripts/main.js
```diff
@@ -1,3 +1,386 @@
...
+const button = document.querySelector('.button');
+const messageStart = document.querySelector('.message-start');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+
+const handlerSlide = e => {
+  switch (e.code) {
+    case 'ArrowLeft':
+      if (slideLeft()) {
+        setTimeout(() => {
```
witflash, line 15: You will have a problem with this approach. If a user will rapidly presses random keys your timeouts will be stacked. 
It's better to block any interaction while animation continues and a tile creates.

It's some improvement but you could think about it.

witflash (approved):
Well done!
You could improve your logic a bit. Check my comment

src/scripts/main.js
oleksusov: > You will have a problem with this approach. If a user will rapidly presses random keys your timeouts will be stacked. It's better to block any interaction while animation continues and a tile creates.
> 
> It's some improvement but you could think about it.

Maybe I should implement slide functions as async functions, wait when all tiles will be moved and then create a new tile.

PR #285, opened 2023-05-01
solution

BipBobBoop:
https://BipBobBoop.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,327 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const container = document.createElement('div');
+
+container.classList.add('container');
+body.insertBefore(container, body.firstChild);
+
+function initializeGame(stage) {
+  gameHeader();
+  gameField();
+  gameNotification(stage);
+}
+
+function gameHeader() {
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,327 @@
...
+    for (let i = column.length - 1; i >= 0; i--) {
+      const cell = column[i];
+
+      cell.classList.remove(`field-cell--${cell.textContent}`);
+
+      const newValue = columnValuesMerged.shift();
+
+      cell.textContent = newValue;
+      cell.classList.add(`field-cell--${newValue}`);
+    }
```
witflash, line 249: Could you try to generalize moving logic and avoid code duplication? 
You make the same things inside every moving function and can make a general method that will take the params it needs. So, potentially your code will be shorter, cleaner, and more readable.

witflash (changes requested):
Great work! 
All logic works fine 👍🏼
But could you improve your code a little bit?
Check my comment, please.

src/scripts/main.js
BipBobBoop: Fixed all.

witflash (approved):
Excellent!
Well done 👍

PR #286, opened 2023-05-01
2048 First commit

dudar-y:
[DEMO LINK](https://dudar-y.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -32,6 +32,12 @@ function addRandomNumber(num) {
   };
 };
 
+function repeater(f, count) {
+  for (let i = 0; i < count; i++) {
+    f();
+  };
+};
+
 function isGameLost() {
   let result = true;
 
@@ -50,7 +56,7 @@ function isGameLost() {
     };
   });
 
-  if (result === true) {
+  if (result) {
     const columns = [];
     const fieldRows = [...gameField.rows];
 
@@ -144,6 +150,9 @@ function moveCells(firsCell, secondCell) {
   const firstValue = Number(firsCell.innerText);
   const secondValue = Number(secondCell.innerText);
 
+  const cellsCompare = firstValue !== 0 && firstValue === secondValue
+    && !blockedCells.includes(firsCell);
+
   if (firstValue !== 0 && secondValue === 0) {
     firsCell.className = 'field-cell';
     firsCell.innerText = '';
@@ -152,10 +161,7 @@ function moveCells(firsCell, secondCell) {
     actionCounter++;
   };
 
-  if (
-    firstValue !== 0 && firstValue === secondValue
-    && !blockedCells.includes(firsCell)
-  ) {
+  if (cellsCompare) {
     firsCell.className = 'field-cell';
     firsCell.innerText = '';
     secondCell.innerText = firstValue * 2;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,379 @@
...
+
+  if (!cell.innerText) {
+    cell.innerText = num;
+    cell.classList.add(`field-cell--${num}`);
+  } else {
+    addRandomNumber(num);
+  };
+};
+
+// Some Function to end game
```
loralevitska, line 37: Remove comments. Fix all cases please

loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements (check comments)

- add winner text when cell have 2048 value
 image
- remove outline 
image

dudar-y:
> 
You have to get this value from the two cells by concatenating them, since the validation happens during the concatenation.
It is wrong?
завантаження

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+  startGame();
+});
+
+document.addEventListener('keydown', (e) => {
+  const freeCells = getFreeCells();
+
+  switch (e.key) {
+    case 'ArrowRight':
+      moveRight();
+      moveRight();
```
Viktor-Kost, line 259: Repeated calls to the moveRight() function can be replaced with a loop or a function that accepts a parameter for the number of times to move.
Apply for similar cases

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+
+    fieldRows.forEach(row => {
+      column.push([...row.cells][i]);
+    });
+
+    columns.push(column);
+  };
+
+  columns.forEach(column => {
+    for (let i = 1; i < 4; i++) {
```
Viktor-Kost, line 216: it's better to create constants to avoid `magic` numbers

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+    firsCell.innerText = '';
+    secondCell.innerText = firstValue;
+    secondCell.classList.add(`field-cell--${firstValue}`);
+    actionCounter++;
+  };
+
+  if (
+    firstValue !== 0 && firstValue === secondValue
+    && !blockedCells.includes(firsCell)
+  ) {
```
Viktor-Kost, line 158: move such a long if condition to a variable for better readability and maintainability.

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+
+      if ((value === nextValue) || value === '') {
+        result = false;
+
+        return;
+      };
+    };
+  });
+
+  if (result === true) {
```
Viktor-Kost, line 53:
```suggestion
  if (result) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,379 @@
...
+function isGameLost() {
+  let result = true;
+
+  [...gameField.rows].forEach(row => {
+    const cells = [...row.cells];
+
+    for (let i = 0; i < cells.length - 1; i++) {
+      if (
+        (cells[i].innerText === cells[i + 1].innerText)
+        || cells[i].innerText === ''
```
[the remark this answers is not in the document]
Viktor-Kost: isn't fixed for some if conditions

Viktor-Kost (changes requested):
Cool!
Some minor fixes left

src/scripts/main.js
dudar-y: Explain in more detail, please...

loralevitska: If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

"magic numbers": `4` is not entirely clear what the number 4 means, the question arises why not 3?, not 7?, why 4?, where did it come from? It is better to put such numbers in a variable and name them so that it is clear from the name what this number means

loralevitska (approved):

PR #296, opened 2023-05-11
v.01

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -27,7 +27,7 @@ const cols = [...field.rows].reduce((res, el) => {
   return res;
 }, []);
 
-startButton.onclick = e => {
+startButton.addEventListener('click', e => {
   if (!startButton.toggle) {
     document.addEventListener('keydown', game);
     showMessage();
@@ -44,7 +44,7 @@ startButton.onclick = e => {
     start();
     showMessage();
   }
-};
+});
 
 function game(e) {
   if (e.key === 'ArrowRight' || e.key === 'ArrowLeft') {
@@ -56,7 +56,7 @@ function game(e) {
   };
 
   addNewNumber(rows);
-  classHandler();
+  cellClassHandler();
 }
 
 function move(arr, direction) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,191 @@
...
+  } else {
+    document.removeEventListener('keydown', game);
+    document.addEventListener('keydown', game);
+    score.innerText = 0;
+    cellList.forEach(el => (el.innerText = ''));
+    start();
+    showMessage();
+  }
+};
+
```
polosanya, line 48: Use `addEventListener` instead of `onclick` to assign an event handler.

src/scripts/main.js
```diff
@@ -1,3 +1,191 @@
...
+  arr.forEach(el => {
+    if (el.innerText) {
+      result.push(el.innerText);
+    }
+  });
+
+  return result;
+};
+
+function classHandler() {
```
polosanya, line 159: Use meaningful function and variable names for better readability.
image

src/scripts/main.js
```diff
@@ -1,3 +1,191 @@
...
+      for (let index = el.length - 1; index >= 0; index--) {
+        if (clear.length) {
+          el[index].innerText = clear.pop();
+        } else {
+          el[index].innerText = '';
+        }
+      }
+    }
+
+    if (direction === 'ArrowLeft' || direction === 'ArrowUp') {
```
polosanya, line 89: It is better to use switch case instead

src/scripts/main.js
```diff
@@ -1,3 +1,191 @@
...
+      for (let ind = 0; ind < el.length; ind++) {
+        if (clear.length) {
+          el[ind].innerText = clear.shift();
+        } else {
+          el[ind].innerText = '';
```
polosanya, line 94: You can use ternary operator here

polosanya (changes requested):
Good job! Let's improve your code a bit)

OleziO (approved):

PR #300, opened 2023-05-13
Solution 20248_game

NazariiAlieksieiev:
[DEMO LINK](https://NazariiAlieksieiev.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -208,24 +208,17 @@ function isCanMove() {
 };
 
 function checkRows() {
-  for (let i = 0; i < 4; i++) {
-    if (newState[i].some(cell => cell === 0)
-      || newState[i]
-        .some((cell, j) => cell === newState[i][j + 1])) {
-      return true;
-    }
-  }
-
-  return false;
+  return newState.some(row =>
+    row.some((cell, index) =>
+      cell === 0 || cell === row[index + 1]
+    )
+  );
 }
 
 function checkColumns() {
-  for (let i = 0; i < 4; i++) {
-    if (newState[i]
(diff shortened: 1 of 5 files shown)


src/styles/main.scss
```diff
@@ -20,9 +20,11 @@ body {
   text-align: center;
   vertical-align: center;
   user-select: none;
+  transition: all 0.5s ease-in-out;
```
Anastasiia-Svintsova, line 23: Consider specifying which properties should be transitioned instead of using 'all' to improve performance.

src/scripts/main.js
```diff
@@ -1,3 +1,231 @@
...
+    return true;
+  }
+
+  tarnsporteState();
+
+  return checkColumns();
+};
+
+function checkRows() {
+  for (let i = 0; i < 4; i++) {
```
Anastasiia-Svintsova, line 211: Use functional programming methods like `some` instead of traditional loops for iterating over arrays.
```
return newState.some(row =>
    row.some((cell, index) =>
      cell === 0 || cell === row[index + 1]
    )
  );

File: src/scripts/main.js [hunk 3]
```diff
@@ -1,3 +1,231 @@
...
+        .some((cell, j) => cell === newState[i][j + 1])) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+function checkColumns() {
+  for (let i = 0; i < 4; i++) {
```
Anastasiia-Svintsova, line 223: Use functional programming methods like `some` instead of traditional loops for iterating over arrays.

Anastasiia-Svintsova (approved):

PR #301, opened 2023-05-14
Develop

MykolaKaradzha:
[DEMO LINK](https://MykolaKaradzha.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,279 @@
 'use strict';
 
-// write your code here
+let board;
+let score = 0;
+const rows = 4;
+const columns = 4;
+let isFirstMove = true;
+let changed = {
+  left: true,
+  right: true,
+  up: true,
+  down: true,
+};
+
+const startMessage = document.querySelector('.message-start');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,279 @@
...
+  });
+};
+
+const setGame = () => {
+  board = [
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+  ];
```
VitaliyBondarenko1982, line 91: consider  to create constant for initial board, as it uses at least two times

VitaliyBondarenko1982 (changes requested):
check such case from screed.
in reference when try to move to top or right from such position shouldn't appear new digit (because existent cells can't to move), but in your solution always appears new cell with digit

src/scripts/main.js
MykolaKaradzha: Thanks, done)

etojeDenys (approved):
lgtm

PR #306, opened 2023-05-17
Develop

IhorChaikovskyi:
for mentors:

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,370 @@
 'use strict';
 
-// write your code here
+const start = document.querySelector('.start');
+const fieldCell = document.querySelectorAll('.field-cell');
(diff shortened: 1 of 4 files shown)


tetiana-ivanchenko (changes requested):
image
I see message "You lose", but I have possible moves

src/scripts/main.js
```diff
@@ -1,3 +1,399 @@
...
+
+document.addEventListener('keydown', (e) => {
+  if (start.textContent === 'Start') {
+    start.classList.add('restart');
+    start.textContent = 'Restart';
+    start.style.fontSize = '18px';
+  }
+
+  switch (e.key) {
+    case 'ArrowDown':
```
OleziO, line 90: Try to move all cases to the object

OleziO (approved):

PR #317, opened 2023-05-27
● 2048 with localStorage & mobileVer

d-hubych:
- [DEMO LINK](https://d-hubych.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/helpers.js
+++ b/src/scripts/helpers.js
@@ -21,30 +21,24 @@ export function checkPossibleMoves(field, size) {
       const center = field[r][c];
       
       if (center !== 0) {
-        console.log(center, 'center')
         const Up = (r > 0) ? field[r - 1][c] : null;
         const Down = (r < size - 1) ? field[r + 1][c] : null;
         const Left = (c > 0) ? field[r][c - 1] : null;
         const Right = (c < size - 1) ? field[r][c + 1] : null;
 
         if (Left === 0 || Left === center) {
-        console.log(Left, 'left')
-
           !possibleMoves.includes('ArrowLeft') && possibleMoves.push('ArrowLeft');
         }
-  
+
         if (Right === 0 || Right === center) {
-          console.log(Right, 'Right')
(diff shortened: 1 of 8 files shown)


src/scripts/helpers.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  return rotatedField;
+};
+
+export function checkPossibleMoves(field, size) {
+  let possibleMoves = [];
+
+  for (let r = 0; r < size; r++) {
+    for (let c = 0; c < size; c++) {
```
VitaliyBondarenko1982, line 16: better to use full word for variables - `row`, `column`
it will increase readibility

src/scripts/helpers.js
```diff
@@ -0,0 +1,66 @@
...
+      if (possibleMoves.length === 4) {
+        break;
+      }
+
+      const center = field[r][c];
+      
+      if (center !== 0) {
```
VitaliyBondarenko1982, line 23:
```suggestion
      if (center) {
```

src/scripts/helpers.js
```diff
@@ -0,0 +1,66 @@
...
+    return;
+  }
+
+  const isVertital = Math.abs(difY) > Math.abs(difX);
+
+  if (isVertital) {
+    return (difY < 0) ? 'ArrowUp' : 'ArrowDown';
+  } else {
+    return (difX < 0) ? 'ArrowLeft' : 'ArrowRight';
+  }
```
VitaliyBondarenko1982, line 65:
```suggestion
  const directionY = difY < 0 ? 'ArrowUp' : 'ArrowDown';
  const directionX = difX < 0 ? 'ArrowLeft' : 'ArrowRight';
  
  return isVertical ? directionY : directionX
```
try to avoid if else where it is possible.
as for me such code more readable

VitaliyBondarenko1982 (approved):
Good job.
Left a few comment about code style.

PR #324, opened 2023-05-31
solution 2048

MargaritaChorniak:
- [DEMO LINK](https://MargaritaChorniak.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,222 @@
 'use strict';
 
-// write your code here
+const button = document.querySelector('button');
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,222 @@
...
+  });
+
+  scoreField.innerText = score;
+}
+
+function move(e) {
+  newState = state;
+
+  switch (e.key) {
+    case 'ArrowLeft':
```
OleziO, line 95: Try to move all cases to the object.

OleziO (approved):

PR #334, opened 2023-06-10
first attempt

batiukova-nataliia:
[DEMO LINK](https://batiukova-nataliia.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,331 @@
 'use strict';
 
-// write your code here
+let board;
+let score = 0;
+const rowsAmount = 4;
+const columnsAmount = 4;
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+
+const rows = document.querySelectorAll('tr');
+
+const buttonStart = document.querySelector('.start');
+
+buttonStart.addEventListener('click', () => {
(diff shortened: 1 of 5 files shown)


VitaliyBondarenko1982 (changes requested):
looks good.
found some bug.
in original game shouldn't appear new numbers, if we don't have any cases to move old numbers or unite same numbers.
looks on screencast
I always press top arrow
chrome-capture-2023-5-10

src/scripts/main.js
```diff
@@ -1,3 +1,354 @@
...
+
+  for (let r = 0; r < rowsAmount; r++) {
+    for (let c = 0; c < columnsAmount; c++) {
+      const tile = document.getElementById(r.toString() + '-' + c.toString());
+
+      if (tile.innerText === '0') {
+        tile.innerText = '';
+      }
+    }
+  }
```
etojeDenys, line 304: we can move this logic to some helper and reuse everywhere

etojeDenys (changes requested):

IvanFesenko (changes requested):
After loose if you have possible moves game will continue - looks like "lose state" is not valid
https://github.com/mate-academy/js_2048_game/assets/61904995/87ae946a-b4af-4afc-a779-d65cf3d190f2

Viktor-Kost (approved):

PR #337, opened 2023-06-11
added solution

sheva10barca:
- [DEMO LINK](https://sheva10barca.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,203 @@
 'use strict';
 
-// write your code here
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+
+const gameField = document.querySelector('.game-field');
+const gameFieldWidth = 4;
+
+const buttonStart = document.querySelector('.start');
+const gameScore = document.querySelector('.game-score');
+
+const winScore = 2048;
+let score = 0;
+
+let gameBoard = boardInit();
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,203 @@
...
+        tile.classList.add('field-cell--' + tileValue);
+      } else {
+        tile.innerText = '';
+      }
+    });
+  });
+}
+
+function duplicateGameBoard() {
+  return gameBoard.map((row) => [...row]);
```
etojeDenys, line 108: Instead of using 'map', you can use 'slice()' method to create a shallow copy of the gameBoard. Replace 'return gameBoard.map((row) => [...row]);' with 'return gameBoard.map((row) => row.slice());'

src/scripts/main.js
```diff
@@ -1,3 +1,203 @@
...
+}
+
+function movesController(e) {
+  makeMove(e.key);
+  fillBoard();
+  handleWin();
+}
+
+buttonStart.addEventListener('click', startGame);
+buttonStart.addEventListener('click', restartGame);
```
etojeDenys, line 203: You are adding both 'startGame' and 'restartGame' event listeners to the same 'click' event. Consider refactoring your code to only have one event listener for the 'click' event and to handle starting and restarting the game within that single function.

etojeDenys (changes requested):
small nits

VitaliyBondarenko1982 (approved):
Well done

PR #341, opened 2023-06-12
Add Solution

dimarogkov:
If we do everything according to the same pattern, then I won't even try to do it myself.
[DEMO LINK](https://dimarogkov.github.io/js_2048_game/)


demetrxx (approved):
Well done 👍 Go ahead!

PR #342, opened 2023-06-12
implementation the game 2048

NadieinOleh:
https://NadieinOleh.github.io/js_2048_game/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -55,7 +55,7 @@
       <div class="message-container">
         <p class="message message-lose hidden">You lose! Restart the game?</p>
         <p class="message message-win hidden">Winner! Congrats! You did it!</p>
-        <p class="message message-start">Press "Start" to begin game. Good luck!</p>
+        <p class="message message-start ">Press "Start" to begin game. Good luck!</p>
       </div>
     </div>
     <script type="text/javascript" src="scripts/main.js"></script>
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,345 @@
 'use strict';
 
-// write your code here
+const start = document.querySelector('.start');
+const scoreDisplay = document.querySelector('.game-score');
+const fieldRowAll = document.querySelectorAll('.field-row');
+const messageLose = document.querySelector('.message-lose');
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+
+let board = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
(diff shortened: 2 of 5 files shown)


lerastarynets (changes requested):
Unfortunately I face the following error when clicking start :(
Screenshot at Jun 12 18-51-27

src/index.html
```diff
@@ -55,7 +55,7 @@ <h1>2048</h1>
       <div class="message-container">
         <p class="message message-lose hidden">You lose! Restart the game?</p>
         <p class="message message-win hidden">Winner! Congrats! You did it!</p>
-        <p class="message message-start">Press "Start" to begin game. Good luck!</p>
+        <p class="message message-start ">Press "Start" to begin game. Good luck!</p>
```
Anastasiia-Svintsova, line 58: There is an extra space in the class attribute, please remove it.

src/scripts/main.js
```diff
@@ -1,3 +1,352 @@
...
+
+  checkAndAction(check);
+}
+
+function moveLeft(check) {
+  for (let i = 0; i < board.length; i++) {
+    board[i] = shiftRowLeft(board[i]);
+  }
+
+  for (let i = 0; i < board.length; i++) {
```
Anastasiia-Svintsova, line 107: Please replace the traditional loop with a functional programming method like 'forEach' for better readability.

src/scripts/main.js
```diff
@@ -1,3 +1,352 @@
...
+    element.classList.add(addClass);
+  }
+}
+
+function checkIfPlayerWon() {
+  const pointForVictory = 2048;
+
+  for (const row of board) {
+    for (const cell of row) {
+      if (cell === pointForVictory) {
```
Anastasiia-Svintsova, line 194: Try to avoid nesting

src/scripts/main.js
```diff
@@ -1,3 +1,352 @@
...
+        changeClassCell(cell);
+      }
+    }
+  }
+
+  scoreDisplay.textContent = score;
+}
+
+function startGame() {
+  board = [
```
Anastasiia-Svintsova, line 294: You create the same board twice. Would be great to create some constant like a defaultBoard and just make copies then.

Anastasiia-Svintsova (approved):

polosanya (commented):
At start there are always only 2 and 2, but would be great to have 4 with 10% chance too
2048-first
Actually, up and down keys don't work correctly. Here is an example of always pressing down key. You can see that it doesn't work as expected
2048-down

polosanya (approved):
👍

PR #343, opened 2023-06-12
added solution

constantweb:
[DEMO LINK](https://constantweb.github.io/js_2048_game/)


demetrxx (changes requested):
1) disable keys press handlers when game is not started
2) fix following bug:

image

Andriy-Shymkiv (approved):
nice work, btw you can try to combine your functions for moving fields into a single one, pass a direction ('up', 'down', 'left', 'right') as a param and configure loops and all the values of which are different, for ex:
```js
function moveField(direction) {
  let index;

  if (direction === 'up') {
     index = someValue;
  }
}
```

PR #350, opened 2023-06-15
game-2048

AnastasiiaVorobets:
- [DEMO LINK](https://AnastasiiaVorobets.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -8,8 +8,8 @@ const loseMessage = document.querySelector('.message-lose');
 const winMessage = document.querySelector('.message-win');
 
 const gameSize = 4;
-let score = 0;
 const winScore = 2048;
+let score = 0;
 
 const gameGrid = [
   [0, 0, 0, 0],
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,305 @@
...
+
+      if (cellValue !== 0) {
+        const cellText = document.createElement('span');
+
+        cellText.textContent = cellValue;
+        cell.appendChild(cellText);
+      }
+    }
+  }
+  score = maxNumber;
```
volodymyr-soltys97, line 106: The score should be calculated based on the sum of all merged cells, not the max cell value. Update the score calculation accordingly.

volodymyr-soltys97 (changes requested):
Good job 🔥 
Let's to fix
Every time I click restart I get a completely blank field with no initial values
https://github.com/mate-academy/js_2048_game/assets/64750958/d3130611-8fa3-4b9b-87ab-eb50212669c0

volodymyr-soltys97 (approved):
Awesome 🚀

PR #351, opened 2023-06-15
create 2048

pos1bl:
[DEMO LINK](https://pos1bl.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,418 @@
 'use strict';
 
-// write your code here
+Array.prototype.random = function() {
+  return this[Math.floor((Math.random() * this.length))];
+};
+
+const game = document.querySelector('.container');
+const startButton = game.querySelector('.button');
+const messages = game.querySelector('.message-container');
+const gameField = game.querySelectorAll('.field-cell');
+const startMessage = messages.querySelector('.message-start');
+const loseMessage = messages.querySelector('.message-lose');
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,418 @@
...
+};
+
+const checkWin = (object) => {
+  const values = Object.values(object);
+
+  return values.some(value => value === finalResult);
+};
+
+const upArrow = (fields) => {
+  for (let i = 4; i <= 7; i++) {
```
nazarmatsevych, line 137: There's a lot of repetitive code for handling arrow key movements. Consider creating a more generic function to handle all cases.

src/scripts/main.js
```diff
@@ -1,3 +1,418 @@
...
+
+    freeFields = Object
+      .keys(Object
+        .fromEntries(Object
+          .entries(fields)
+          .filter(x => x[1] === '')));
+
+    paint(gameField);
+
+    score.innerHTML = Object.values(fields).reduce((a, b) => +a + +b);
```
nazarmatsevych, line 100: The score should be increased based on the sum of merged cells, not the sum of all cells. Adjust the logic to calculate score correctly.

nazarmatsevych (changes requested):
Good work but let's improve a bit.
Every time I click restart I get a completely blank field with no initial values so it doesn't match the working example
image
image

PR #357, opened 2023-06-17
start

vitalii-fedusov:
- [DEMO LINK](https://vitalii-fedusov.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,105 @@
 'use strict';
 
-// write your code here
+function getRandonCell() {
+  let randomCellIndex;
+
+  do {
+    randomCellIndex = Math.floor(Math.random() * 16);
+  } while (fields[randomCellIndex].classList.length > 1);
+
+  return randomCellIndex;
+}
+
+function appearingTwoCells() {
+  const firstIndex = getRandonCell();
+  const firstField = fields[firstIndex];
+
+  firstField.textContent = `${number1}`;
+
+  if (number1 === '2') {
+    firstField.classList.add('field-cell--2');
+  } else {
+    firstField.classList.add('field-cell--4');
+  }
+
+  const secondIndex = getRandonCell();
+  const secondField = fields[secondIndex];
+
+  secondField.textContent = `${number2}`;
+
+  if (number2 === '2') {
+    secondField.classList.add('field-cell--2');
+  } else {
+    secondField.classList.add('field-cell--4');
+  }
+}
+
+const fields = document.querySelectorAll('.field-cell');
+const startButton = document.querySelector('.button');
+
+const arrOfFirstNumbers = ['2', '2', '2', '2', '2', '4', '2', '2', '2', '2'];
+const randomIndex1 = Math.floor(Math.random() * 10);
+const randomIndex2 = Math.floor(Math.random() * 10);
+const number1 = arrOfFirstNumbers[randomIndex1];
+const number2 = arrOfFirstNumbers[randomIndex2];
+
+startButton.addEventListener('click', () => {
+  startButton.textContent = 'restart';
+  startButton.classList.add('restart');
+
+  appearingTwoCells();
+});
+
+// addEventListener('keydown', (event) => {
+//   if (event.code === 'ArrowLeft') {
+//     console.log('left');
+//   }
+
+//   if (event.code === 'ArrowRight') {
+//     console.log('right');
+//   }
+
+//   if (event.code === 'ArrowUp') {
+//     console.log('up');
+//   }
+
+//   if (event.code === 'ArrowDown') {
+//     console.log('down');
+//   }
+// });
+
+// const table = [
+//   [0, 0, 0, 0],
+//   [0, 0, 0, 0],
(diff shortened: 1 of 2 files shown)


VitaliyBondarenko1982 (changes requested):
I have wriiten answer on your question about how to move cells in fe_chat

src/scripts/main.js
```diff
@@ -1,3 +1,403 @@
...
+    restart();
+    updateGameFields();
+    // console.log(table);
+  }
+});
+
+function getRandonCell() {
+  const arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+
+  for (let i = 15; i >= 0; i--) {
```
VitaliyBondarenko1982, line 214: lloks like redundant to make such hardcoded arr.
better to iterate through table, filter empty cells
and than randomly choose one,
can be something like that
```
function getRandonCell() {
const emptyCells = [];

  table.forEach((row, rowIndex) => {
    row.forEach((cell, columnIndex) => {
      if (cell === 0) {
        emptyCells.push([rowIndex, columnIndex]);
      }
    });
  });

  return emptyCells[Math.floor(Math.random() * emptyCells.length)];
}

**Review by VitaliyBondarenko1982 — CHANGES_REQUESTED** [review]
Looks good
Needs some improvements.
- no need to show zeros for empty cells
Screenshot from 2023-07-30 17-03-53
- after restart should immediately start new game

File: src/scripts/main.js [hunk 2]
```diff
@@ -1,3 +1,449 @@
...
+    const copyTableValue = copyTable[rowIndex][colIndex];
+
+    if (tableValue !== copyTableValue) {
+      appearingOneRandomCell();
+      updateGameFields();
+
+      break;
+    }
+  }
+}
```
VitaliyBondarenko1982, line 366: almost same code. difference just in up or down function. combine it in one function which will be accept function up or down as argument

src/scripts/main.js
```diff
@@ -1,3 +1,449 @@
...
```
VitaliyBondarenko1982, line 318: for moveLeft and moveRight also almost same code and can be combined in one function.
or at least loop same for four move functions and can be writed in separate function

VitaliyBondarenko1982 (changes requested):
Looks good.
Almost done.
Consider some additional refactoring.

src/scripts/main.js
```diff
@@ -1,3 +1,394 @@
...
+  const mergedArr = arr.filter(item => item !== 0);
+
+  while (mergedArr.length < 4) {
+    mergedArr.unshift(0);
+  }
+
+  table[0][index] = mergedArr[0] || 0;
+  table[1][index] = mergedArr[1] || 0;
+  table[2][index] = mergedArr[2] || 0;
+  table[3][index] = mergedArr[3] || 0;
```
Viktor-Kost, line 57: for (let i = 0; i < table.length; i++) {
      table[i][index] = mergedArr[i] || 0;
    }

src/scripts/main.js
```diff
@@ -1,3 +1,394 @@
...
+
+      break;
+    }
+  }
+}
+
+document.addEventListener('keydown', eventParam => {
+  if (eventParam.key === 'ArrowLeft') {
+    moveHorizontally(left);
+    loseCheck();
```
Viktor-Kost, line 319: I'd recommend you to optimize your `loseCheck`. First check if noEmptyCell and only then, if it's true, check noCellToMerge. Since now your start two extra loops inside noCellToMerge each user step

Also, you can invoke this function once after `if` conditions instead of invoking for each if condition

src/scripts/main.js
```diff
@@ -1,3 +1,394 @@
...
+let touchStartX = 0;
+let touchStartY = 0;
+
+document.addEventListener('touchmove', (eventPar) => {
+  eventPar.preventDefault();
+});
+
+gameTable.addEventListener('touchmove', (eventPar) => {
+  eventPar.preventDefault();
+});
```
Viktor-Kost, line 349: usually we name either event or just e
```suggestion
document.addEventListener('touchmove', (event) => {
  event.preventDefault();
});

gameTable.addEventListener('touchmove', (event) => {
  event.preventDefault();
});
```

src/scripts/main.js
```diff
@@ -1,3 +1,394 @@
...
+
+gameTable.addEventListener('touchend', (eventPar) => {
+  const touchEndX = eventPar.changedTouches[0].clientX;
+  const touchEndY = eventPar.changedTouches[0].clientY;
+  const touchDiffX = touchEndX - touchStartX;
+  const touchDiffY = touchEndY - touchStartY;
+
+  if (Math.abs(touchDiffX) > Math.abs(touchDiffY)) {
+    if (touchDiffX > 50) {
```
Viktor-Kost, line 363: I'd recommend you not using magic number. Create variables for better code apprehension

Viktor-Kost (changes requested):
Nice!
Some moments can be improved

IvanFesenko (approved):
Well done

PR #358, opened 2023-06-17
solution

drhtka:
https://drhtka.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,224 @@
 'use strict';
 
-// write your code here
+const button = document.querySelector('button');
+
+const messageStart = document.querySelector('.message-start');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+const field = document.querySelector('tbody');
+const scoreField = document.querySelector('.game-score');
+
+let state = [
+  [0, 0, 0, 0],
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,224 @@
...
+  });
+
+  scoreField.innerText = score;
+}
+
+function move(e) {
+  newState = state;
+
+  switch (e.key) {
+    case 'ArrowLeft':
```
OleziO, line 97: Move to the object. Example of usage: case Directions.ArrowLeft etc.

OleziO (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,223 @@
...
+      }
+    });
+  });
+
+  scoreField.innerText = score;
+}
+
+function move(e) {
+  const Directions = {
+    ArrowLeft: left,
```
TarasHoliuk, line 95:
```suggestion
    ArrowLeft: 'left',
```
and use it in the switch as:
```
case Directions.ArrowLeft:
  left();
```
currently it doesn't work

TarasHoliuk (changes requested):

SanyaBratashchuk (approved):

PR #367, opened 2023-06-22
solution

JOWISSA:
[DEMO LINK](https://JOWISSA.github.io/js_2048_game/)


demetrxx (changes requested):
image
Seems like styles don't work properly and player should not see any numbers before game starts

sTorba24 (changes requested):
Check styles, cells should be colorized

also score calculating is weird, it's not like on original game

https://play2048.co/

sTorba24 (approved):
Much better :) Approved

PR #376, opened 2023-06-23
add task solution 2048 game

antonina-klishch:
https://antonina-klishch.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,248 @@
 'use strict';
 
-// write your code here
+let valueBoard;
+const rows = 4;
+const columns = 4;
+let score = 0;
+
+const buttonStart = document.querySelector('.start');
+const fieldRows = document.querySelectorAll('tr');
+const cells = document.querySelectorAll('td');
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+
+const board = [];
+
+[...fieldRows].forEach(row => {
+  const cell = row.querySelectorAll('td');
+
+  board.push(cell);
+});
+
+function setGame() {
+  valueBoard = [
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
+    [0, 0, 0, 0],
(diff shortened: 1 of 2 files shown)


etojeDenys (changes requested):
if no cases to move or merge cell shouldn't appear new 2/4 numbers
on start screencast I pressed up arrow and appear 2 number, but shouldn't

https://github.com/mate-academy/js_2048_game/assets/70220667/175fa332-d35e-4464-b96c-a81b140048a0

antonina-klishch:
> 

I didn't understand why the number shouldn't appear during shooting. You pressed button array up (it case for moving) and a number appeared

TarasHoliuk:
The case Denys meant:
now there is only one empty tile, and no one tile will move/merge on Left Arrow click:
image

After pressing Left Arrow `2` appeared in the empty tile, but shouldn't (since nothing changed)

src/scripts/main.js
```diff
@@ -1,3 +1,248 @@
...
+    buttonStart.textContent = 'Restart';
+    buttonStart.classList.add(`restart`);
+    buttonStart.classList.remove(`start`);
+    setGame();
+
+    document.addEventListener('keydown', (e) => {
+      switch (e.key) {
+        case 'ArrowRight':
+          moveRight();
+          addNumber();
```
TarasHoliuk, line 104: you should check whether you really need to add a new number

src/scripts/main.js
```diff
@@ -1,3 +1,248 @@
...
+        default:
+          return;
+      }
+
+      [...cells].forEach(cell => {
+        if (cell.textContent === '2048') {
+          messageWin.classList.remove(`hidden`);
+
+          setTimeout(() => {
+            document.location.reload();
```
TarasHoliuk, line 127: _not critical_. not sure that the desired behavior - you may remove the event listener to prevent the next moves, etc
up to you whether to change it or not

TarasHoliuk (changes requested):

loralevitska (approved):
Good job!

PR #377, opened 2023-06-23
add task solution

Sergej161995:
[DEMO LINK](https://Sergej161995.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,219 @@
 'use strict';
 
-// write your code here
+const score = document.querySelector('.game-score');
+const button = document.querySelector('.button');
+const cells = document.querySelector('tbody');
+const startGameText = document.querySelector('.message-start');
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,219 @@
...
+      }
+    }
+  }
+
+  if (loseGame()) {
+    loseGameText.classList.remove('hidden');
+  }
+}
+
+document.addEventListener('keyup', (e) => {
```
TarasHoliuk, line 121: arrow buttons work before I click "Start":
image

add this listener on the "Start" button click (i.e. inside another listener). Define the function separately that you use here

TarasHoliuk (changes requested):
Well done :fire:
Let's fix one thing

TarasHoliuk (approved):
:+1:

PR #380, opened 2023-06-24
message

Pa1eOrc:
[DEMO LINK](https://Pa1eOrc.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,256 @@
 'use strict';
 
-// write your code here
+const gameContainer = document.querySelector('.container');
+const startButton = gameContainer.querySelector('.button');
+const cells = gameContainer.querySelectorAll('.field-cell');
+const messageStart = gameContainer.querySelector('.message-start');
+const messageWin = gameContainer.querySelector('.message-win');
+const messageLose = gameContainer.querySelector('.message-lose');
+const gameScore = gameContainer.querySelector('.game-score');
+const gameRow = gameContainer.querySelectorAll('tr');
+let movePossible = false;
+
+function generateRandom() {
+  const randomI = Math.floor(Math.random() * cells.length);
+
+  if (cells[randomI].textContent === '') {
+    const randomNum = Math.random() < 0.1 ? '4' : '2';
+
+    cells[randomI].textContent = randomNum;
+    cells[randomI].classList.add('field-cell--' + randomNum);
+  } else {
+    generateRandom();
+  }
+}
+
+startButton.addEventListener('click', () => {
+  if (startButton.classList.contains('start')) {
+    generateRandom();
+    generateRandom();
+    messageStart.classList.add('hidden');
+    startButton.classList.remove('start');
+    startButton.classList.add('restart');
+    startButton.textContent = 'Restart';
+  } else if (startButton.classList.contains('restart')) {
+    cells.forEach((cell) => {
+      cell.textContent = '';
+      cell.className = 'field-cell';
+    });
+    messageStart.classList.remove('hidden');
+    startButton.classList.remove('restart');
+    startButton.classList.add('start');
+    startButton.textContent = 'Start';
+    gameScore.textContent = '0';
+    messageLose.classList.add('hidden');
+    messageWin.classList.add('hidden');
+  }
+});
+
+function updateCells() {
+  cells.forEach((cell) => {
+    if (cell.textContent === '') {
+      cell.className = 'field-cell';
+    } else {
+      cell.className = 'field-cell';
+      cell.classList.add('field-cell--' + cell.textContent);
+    }
+  });
+}
+
+document.addEventListener('keydown', (e) => {
+  if (e.key === 'ArrowLeft') {
+    slideLeft();
+    checkGameOver();
+  }
+
+  if (e.key === 'ArrowRight') {
+    slideRight();
+    checkGameOver();
+  }
+
+  if (e.key === 'ArrowUp') {
+    slideUp();
+    checkGameOver();
+  }
+
+  if (e.key === 'ArrowDown') {
+    slideDown();
+    checkGameOver();
+  }
+
+  checkGameWin();
+});
+
+function filterZero(cellsArray) {
+  return cellsArray.filter(i => i !== '');
+}
+
+function slide(cellsArray) {
+  let arrayToSlide = filterZero(cellsArray);
+  let updatedScore = 0;
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,256 @@
...
+const messageLose = gameContainer.querySelector('.message-lose');
+const gameScore = gameContainer.querySelector('.game-score');
+const gameRow = gameContainer.querySelectorAll('tr');
+let movePossible = false;
+
+function generateRandom() {
+  const randomI = Math.floor(Math.random() * cells.length);
+
+  if (cells[randomI].textContent === '') {
+    const randomNum = Math.random() < 0.1 ? '4' : '2';
```
VitaliyBodnar, line 17: 0.1, '4', '2' are 'magic numbers', so please move these to constants

src/scripts/main.js
```diff
@@ -1,3 +1,256 @@
...
+  }
+});
+
+function updateCells() {
+  cells.forEach((cell) => {
+    if (cell.textContent === '') {
+      cell.className = 'field-cell';
+    } else {
+      cell.className = 'field-cell';
+      cell.classList.add('field-cell--' + cell.textContent);
```
VitaliyBodnar, line 55:
```suggestion
      cell.className = `field-cell field-cell--${cell.textContent}`;
```

src/scripts/main.js
```diff
@@ -1,3 +1,256 @@
...
+
+  if (e.key === 'ArrowUp') {
+    slideUp();
+    checkGameOver();
+  }
+
+  if (e.key === 'ArrowDown') {
+    slideDown();
+    checkGameOver();
+  }
```
VitaliyBodnar, line 79: - move 'ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown' to the constants 
- please use switch case hare 
- function checkGameOver(), try to call this function only one time, but with the condition

src/scripts/main.js
```diff
@@ -1,3 +1,256 @@
...
+    }
+  }
+
+  if (!emptyCells && !adjacentCells) {
+    messageLose.classList.remove('hidden');
+  }
+}
+
+function checkGameWin() {
+  const winCell = Array.from(cells).some(cell => cell.textContent === '2048');
```
VitaliyBodnar, line 251: move '2048' to the const

VitaliyBodnar (changes requested):
Good Job🔥, let’s make a few changes.

- when the game is finished you need to show the message, like 'You lose! Restart the game?'
Screenshot 2023-06-24 at 16 42 05

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+  return cellsArray.filter(i => i !== '');
+}
+
+function slide(cellsArray) {
+  let arrayToSlide = filterZero(cellsArray);
+  let updatedScore = 0;
+
+  for (let i = 0; i < arrayToSlide.length; i++) {
+    if (arrayToSlide[i] === arrayToSlide[i + 1]) {
+      arrayToSlide[i] *= 2;
```
SerhiiKirik, line 102: You are reassigning an array value directly. This can lead to mutation of the original array. It's a better practice to create a new array with the modified elements.

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+
+function filterZero(cellsArray) {
+  return cellsArray.filter(i => i !== '');
+}
+
+function slide(cellsArray) {
+  let arrayToSlide = filterZero(cellsArray);
+  let updatedScore = 0;
+
+  for (let i = 0; i < arrayToSlide.length; i++) {
```
SerhiiKirik, line 100: Avoid using a for loop for array manipulation like summing, mapping, filtering etc. Here, you are using it for modifying an array. You can use `map`, `forEach` or `reduce` to modify the array instead.

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+const moveLeft = 'ArrowLeft';
+const moveRight = 'ArrowRight';
+const moveUp = 'ArrowUp';
+const moveDown = 'ArrowDown';
+const winnerNumber = '2048';
+let movePossible = false;
+let adjacentCells = false;
+
+function generateRandom() {
+  const randomI = Math.floor(Math.random() * cells.length);
```
SerhiiKirik, line 23: Recursive function 'generateRandom' can lead to maximum call stack size exceeded error if all cells are filled. I suggest you to first check if there is an available space before calling the function and then generate a random number for an empty cell.

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+
+startButton.addEventListener('click', () => {
+  if (startButton.classList.contains('start')) {
+    generateRandom();
+    generateRandom();
+    messageStart.classList.add('hidden');
+    startButton.classList.remove('start');
+    startButton.classList.add('restart');
+    startButton.textContent = 'Restart';
+  } else if (startButton.classList.contains('restart')) {
```
SerhiiKirik, line 44: try to avoid `if / else if` construction, just use `if` here

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+    columnIndex < gameRow[0].childElementCount;
+    columnIndex++
+  ) {
+    const columnCells = Array.from(gameRow).map(
+      row => row.children[columnIndex].textContent);
+
+    const updatedArray = slide(columnCells);
+
+    for (let rowIndex = 0; rowIndex < gameRow.length; rowIndex++) {
+      if (gameRow[rowIndex].children[columnIndex].textContent
```
SerhiiKirik, line 182: split it with variables

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+    cellsArray.forEach((currentCell, i) => {
+      const rightCell = cellsArray[i + 1];
+      const bottomCell = rowIndex < gameRow.length - 1
+        ? gameRow[rowIndex + 1].children[i].textContent
+        : null;
+
+      if ((rightCell && currentCell === rightCell)
+        || (bottomCell && currentCell === bottomCell)) {
+        adjacentCells = true;
+      }
```
SerhiiKirik, line 259: split it with variables

SerhiiKirik (changes requested):
Great 🔥 
Let's try to make your code more cleaner 🧑‍💻

SerhiiKirik (approved):
Nice 🔥

PR #381, opened 2023-06-24
solution

cebanoleksandr:
- DEMO LINK - https://cebanoleksandr.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,269 @@
 'use strict';
 
-// write your code here
+const gameField = document.querySelector('.game-field');
+const cells = document.querySelectorAll('.field-cell');
+const scoreDisplay = document.querySelector('.game-score');
+const startButton = document.querySelector('.button.start');
+const loseMessage = document.querySelector('.message-lose');
+const winMessage = document.querySelector('.message-win');
+const startMessage = document.querySelector('.message-start');
+const tryAgain = document.querySelector('.btn');
+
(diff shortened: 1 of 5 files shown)


YegorKorenkov (changes requested):
I press 'Try again' button but the game won't restart
image

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
 'use strict';
 
-// write your code here
+const gameField = document.querySelector('.game-field');
```
tetiana-ivanchenko, line 3: 'gameField' is assigned a value but never used

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+
+    newBoard[row] = newArr;
+    newArr = [];
+  }
+
+  board = newBoard;
+}
+
+function handleKeyPress(event) {
+  if (event.key === 'ArrowLeft') {
```
tetiana-ivanchenko, line 250: good case to use switch/case

tetiana-ivanchenko (changes requested):

anstsot (changes requested):
Good work, but there is one small bug to fix. After retrying the game started with 2 filled cells instead of one

https://github.com/mate-academy/js_2048_game/assets/22824431/098b626b-06dd-4f01-be9f-12b7f5a25a88

mgrinko:
А де анімація?

PR #388, opened 2023-06-27
2048_game

AlexRaven93:
- [DEMO LINK](https://AlexRaven93.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+}
+
+function cellUp() {
+  for (const column of columns) {
+    rotateCells([...column].reverse());
+  }
+}
+
+function cellDown() {
+  for (const column of columns) {
```
SerhiiKirik, line 83: Avoid using a `for...of` loop to calculate the sum of an array. Instead, use `forEach`

src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+  });
+
+  randomCellFill();
+  randomCellFill();
+};
+
+function isMergePossible() {
+  const wholeField = [...cellsInRow, ...columns];
+
+  for (const line of wholeField) {
```
SerhiiKirik, line 67: Avoid using a `for...of` loop to calculate the sum of an array. Instead, use `forEach` or `reduce`.

src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+
+  if (value === 2048) {
+    winResult.classList.remove('hidden');
+  }
+}
+
+function rotateCells(line) {
+  for (let i = 3; i > 0; i--) {
+    const moveAllowed = clearCells.includes(line[i])
+      && !clearCells.includes(line[i - 1]);
```
SerhiiKirik, line 157: move `line[i - 1]` to the variable

src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+  clearCells.splice(clearCells.indexOf(curr), 1);
+  deleteElement(prev);
+}
+
+function cellMerge(curr, prev) {
+  const value = curr.innerText * 2;
+
+  curr.innerText = value;
+  curr.className = `field-cell field-cell--${value}`;
+  gameScore.innerText = +gameScore.innerText + value;
```
SerhiiKirik, line 143: use `Number` instead `+`
```suggestion
  gameScore.innerText = Number(gameScore.innerText) + value;
```

SerhiiKirik (approved):
Well done 🔥

src/scripts/main.js
SerhiiKirik: do it in all cases

PR #389, opened 2023-06-28
2048

IvanFront-end:
[DEMO LINK](https://IvanFront-end.github.io/js_2048_game/)


sTorba24 (changes requested):
Almost good!

make board visible before start game

check scoring, it should add +16 if you connected 8 and 8? but not 8 points

demetrxx (approved):
Good job! But let's improve your code a bit.

PR #390, opened 2023-06-28
Finished task

artkas03:
[DEMO LINK](https://artkas03.github.io/js_2048_game/)


anstsot (changes requested):
Good work but I've found the bug. The game is over even if there are possible moves
image

mgrinko:
А де анімація?

PR #392, opened 2023-06-30
solution

artem-ivashchenko:
[DEMO LINK](https://artem-ivashchenko.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,258 @@
 'use strict';
 
-// write your code here
+const boardSize = 4;
+
+const boardBody = document.querySelector('.game-field').firstElementChild;
+const score = document.querySelector('.game-score');
+const gameButton = document.querySelector('.start');
+const cells = document.querySelectorAll('td');
+
+let board = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
+
+for (let r = 0; r < boardSize; r++) {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,258 @@
...
+      cell.innerText = '';
+      cell.classList.value = 'field-cell';
+    });
+
+    board = [
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+    ];
```
sTorba24, line 64: you already have board, try to avoid duplicating, i think it can be done in other way

sTorba24 (changes requested):
Good work! Lets improve it

Don't looks so :)
image

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+function start(e) {
+  if (e.key === 'ArrowLeft' || e.key === 'ArrowRight'
+      || e.key === 'ArrowUp' || e.key === 'ArrowDown') {
+    gameButton.classList.value = 'button restart';
+    gameButton.innerText = 'Restart';
+    gameButton.addEventListener('click', gameClick);
+  }
+
+  switch (e.key) {
+    case 'ArrowLeft':
```
polosanya, line 79: It would be great to create variables for keys in order not to hardcode them. Also, you would be able to change the movement keys easily

polosanya (approved):
Good job!

PR #395, opened 2023-06-30
add task solution

VolodymyrPst:
https://volodymyrpst.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,244 @@
 'use strict';
 
-// write your code here
+const startBtn = document.querySelector('.button');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+const messageStart = document.querySelector('.message-start');
+const score = document.querySelector('.game-score');
+const gameScore = document.querySelector('.game-score');
+const rows = document.querySelectorAll('.field-row');
+
+let arrOfNumbers = [
+  ['', '', '', ''],
+  ['', '', '', ''],
+  ['', '', '', ''],
+  ['', '', '', ''],
+];
+let moves = 0;
+let maxNum = 0;
+
+score.innerText = maxNum;
+
+startBtn.addEventListener('click', (e) => {
+  e.target.classList.toggle('restart');
+
+  if (e.target.textContent === 'Start') {
+    e.target.textContent = 'Restart';
+    maxNum = 0;
+
+    arrOfNumbers = [
+      ['', '', '', ''],
+      ['', '', '', ''],
+      ['', '', '', ''],
+      ['', '', '', ''],
+    ];
+    randomNum();
+    paint();
+  } else {
+    e.target.textContent = 'Start';
+  }
+});
+
+if (maxNum >= 2048) {
+  messageWin.classList.toggle('hidden');
+  messageStart.classList.toggle('hidden');
+}
+
+function hasEmpty() {
+  for (let i = 0; i < 4; i++) {
+    for (let j = 0; j < 4; j++) {
+      if (arrOfNumbers[i][j] === '') {
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+function maxValue() {
+  for (const item of arrOfNumbers) {
+    for (const num of item) {
+      if (num !== '' && maxNum < num) {
+        maxNum = num;
+        messageStart.classList.toggle('hidden');
+      }
+    }
+  }
+
+  gameScore.innerText = maxNum;
+}
+
+function randomNum() {
+  if (!hasEmpty()) {
+    messageLose.classList.toggle('hidden');
+    messageStart.classList.toggle('hidden');
+
+    return;
+  }
+
+  let x;
+  let y;
+  const value = 2;
+
+  if (moves === 0) {
+    x = Math.floor(Math.random() * 4);
+    y = Math.floor(Math.random() * 4);
+
+    const x1 = Math.floor(Math.random() * 4);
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,244 @@
...
+let arrOfNumbers = [
+  ['', '', '', ''],
+  ['', '', '', ''],
+  ['', '', '', ''],
+  ['', '', '', ''],
+];
+let moves = 0;
+let maxNum = 0;
+
+score.innerText = maxNum;
```
Viktor-Kost, line 20: The initial score is set to maxNum which is 0. According to task requirements, the score should increase with each move, based on the sum of all merged cells.

src/scripts/main.js
```diff
@@ -1,3 +1,244 @@
...
+
+  return clearRow;
+};
+
+function slide(row) {
+  let clearRow = filterZero(row);
+
+  for (let i = 0; i < clearRow.length - 1; i++) {
+    if (clearRow[i] === clearRow[i + 1]) {
+      clearRow[i] *= 2;
```
Viktor-Kost, line 137: The score is not being updated when cells are merged. You should add the value of the merged cells to the score.

src/scripts/main.js
```diff
@@ -1,3 +1,244 @@
...
+  if (!hasEmpty()) {
+    messageLose.classList.toggle('hidden');
+    messageStart.classList.toggle('hidden');
+
+    return;
+  }
+
+  let x;
+  let y;
+  const value = 2;
```
Viktor-Kost, line 82: The variable 'value' is always 2. According to task requirements, after each move, a 2 or 4 should appear in a random empty cell, with 4 having a 10% probability.

src/scripts/main.js
```diff
@@ -1,3 +1,244 @@
...
+      ['', '', '', ''],
+    ];
+    randomNum();
+    paint();
+  } else {
+    e.target.textContent = 'Start';
+  }
+});
+
+if (maxNum >= 2048) {
```
Viktor-Kost, line 42: The check for maxNum >= 2048 should be inside the game loop or a function that is called during the game loop, not at the top level of the script.

src/scripts/main.js
```diff
@@ -1,3 +1,244 @@
...
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+function maxValue() {
+  for (const item of arrOfNumbers) {
```
Viktor-Kost, line 60: Avoid using a 'for...of' loop to calculate the maximum value in an array. Instead, you can use Math.max() with the spread operator.

Viktor-Kost (changes requested):
Hello!)
Please check the logic of restarting the game. It should reset current values immediately
And check the text below the game. It appears and disappear time to

https://github.com/mate-academy/js_2048_game/assets/117587463/bca703a2-1ce6-4c74-b2b3-5d1fb81416d8

src/scripts/main.js
```diff
@@ -1,3 +1,253 @@
 'use strict';
 
-// write your code here
+const startBtn = document.querySelector('.button');
+// const messageLose = document.querySelector('.message-lose');
```
OleziO, line 4: Remove all comments from the code.

src/scripts/main.js
```diff
@@ -1,3 +1,253 @@
...
+
+  while (clearRow.length < 4) {
+    clearRow.push('');
+  }
+
+  return clearRow;
+}
+
+document.addEventListener('keyup', (e) => {
+  if (e.code === 'ArrowLeft') {
```
OleziO, line 165: Move all directions to the object.

Example of usage: 
`if (e.code === Directions.ArrowLeft)`

OleziO (changes requested):

volodymyr-soltys97 (approved):
Nice work 🚀

anastasiiavorobiova (changes requested):
Fantastic work!
Screenshot 2023-08-17 at 23 45 38
The loss message doesn't disappear after restarting

sTorba24 (changes requested):
Good work!

I already started it :)
image

cells appears even if nothing to move or connect.
to reproduce try to infinite click on same direction - 'right' for example
and then check the original game.
https://2048game.com/

lerastarynets (changes requested):
Why congrats appears after reaching 128? 😅 This also remains if user lose
Screenshot at Aug 19 13-46-59
Screenshot at Aug 19 13-49-48

src/scripts/main.js
```diff
@@ -1,3 +1,293 @@
...
+
+    row.reverse();
+
+    const newRow = slide(row);
+
+    newRow.reverse();
+
+    arrOfNumbers[i] = newRow;
+  }
+  // randomNum();
```
volodymyr-soltys97, line 225: The same as above

src/scripts/main.js
```diff
@@ -1,3 +1,293 @@
...
+
+    newRow.reverse();
+
+    arrOfNumbers[0][i] = newRow[0];
+    arrOfNumbers[1][i] = newRow[1];
+    arrOfNumbers[2][i] = newRow[2];
+    arrOfNumbers[3][i] = newRow[3];
+  }
+
+  // randomNum();
```
volodymyr-soltys97, line 271:
```suggestion
```

volodymyr-soltys97 (changes requested):

volodymyr-soltys97 (approved):
Well done 🔥

PR #397, opened 2023-07-04
Develop

reseneweb:
https://reseneweb.github.io/js_2048_game/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,977 @@
 'use strict';
 
-// write your code here
+const page = document.querySelector('.page');
+const switcherStart = document.querySelector('.start');
+const switcherScore = document.querySelector('.game-score');
+
+let workingArray = Array(16).fill(null);
+let changeIndicator = true;
+let scoreCounter = 0;
+
+function generateInitialValues() {
+  let positionX;
+  let positionY;
+  const startValueX = Math.floor(Math.random() * 9);
+  const startValueY = Math.floor(Math.random() * 9);
+
+  do {
+    positionX = Math.floor(Math.random() * 16);
+    positionY = Math.floor(Math.random() * 16);
+  } while (positionX === positionY);
+
+  if (startValueX !== 0) {
+    page.querySelectorAll('.field-cell')[positionX]
+      .classList.add('field-cell--2');
+    workingArray[positionX] = 2;
+  } else {
+    page.querySelectorAll('.field-cell')[positionX]
+      .classList.add('field-cell--4');
+    workingArray[positionX] = 4;
+  }
+
+  if (startValueY !== 0) {
+    page.querySelectorAll('.field-cell')[positionY]
+      .classList.add('field-cell--2');
+    workingArray[positionY] = 2;
+  } else {
+    page.querySelectorAll('.field-cell')[positionY]
+      .classList.add('field-cell--4');
+    workingArray[positionY] = 4;
+  };
+
+  switcherStart.classList.remove('start');
+  switcherStart.classList.add('restart');
+  switcherStart.textContent = 'Restart';
+  page.querySelector('.message-start').classList.add('hidden');
+}
+
+function generateTwoOrFour() {
+  let findAllIndices = [];
+
+  findAllIndices = [];
+
+  for (let i = 0; i < workingArray.length; i++) {
+    if (workingArray[i] !== null) {
+      findAllIndices.push(i);
+    }
+  }
+
+  let positionX;
+  const startValueX = Math.floor(Math.random() * 9);
+
+  do {
+    positionX = Math.floor(Math.random() * 16);
+  } while (findAllIndices.includes(positionX));
+
+  if (startValueX !== 0) {
+    page.querySelectorAll('.field-cell')[positionX]
+      .classList.add('field-cell--2');
+    workingArray[positionX] = 2;
+  } else {
+    page.querySelectorAll('.field-cell')[positionX]
+      .classList.add('field-cell--4');
+    workingArray[positionX] = 4;
+  }
+};
+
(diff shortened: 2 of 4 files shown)


src/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="page">
```
lerastarynets, line 2: The 'class' attribute added to the 'html' tag is not recommended. The 'class' is usually applied to 'body' or any specific html element inside 'body'.

src/scripts/main.js
```diff
@@ -1,3 +1,1004 @@
...
+  switcherStart.classList.remove('start');
+  switcherStart.classList.add('restart');
+  switcherStart.textContent = 'Restart';
+  page.querySelector('.message-start').classList.add('hidden');
+}
+
+function generateTwoOrFour() {
+  let findAllIndices = [];
+
+  findAllIndices = [];
```
lerastarynets, line 52: why do you reassign `findAllIndices` to an empty array once again?

src/scripts/main.js
```diff
@@ -1,3 +1,1004 @@
...
+    ch.classList.remove('field-cell--4');
+    ch.classList.remove('field-cell--8');
+    ch.classList.remove('field-cell--16');
+    ch.classList.remove('field-cell--32');
+    ch.classList.remove('field-cell--64');
+    ch.classList.remove('field-cell--128');
+    ch.classList.remove('field-cell--256');
+    ch.classList.remove('field-cell--512');
+    ch.classList.remove('field-cell--1024');
+    ch.classList.remove('field-cell--2048');
```
lerastarynets, line 98: I think it's better to use array iteration methods for such cases

src/scripts/main.js
```diff
@@ -1,3 +1,1004 @@
...
+function upwardMovement() {
+  if (disableFunctions) {
+    return;
+  }
+
+  changeIndicator = false;
+
+  let allIndices = [];
+
+  allIndices = [];
```
lerastarynets, line 142: why do you reassign allIndices to an empty array once again?

src/scripts/main.js
```diff
@@ -1,3 +1,1004 @@
...
+  page.querySelector('.message-start').classList.add('hidden');
+}
+
+function generateTwoOrFour() {
+  let findAllIndices = [];
+
+  findAllIndices = [];
+
+  for (let i = 0; i < workingArray.length; i++) {
+    if (workingArray[i] !== null) {
```
lerastarynets, line 55: pls [see falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
```suggestion
    if (workingArray[i]) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,1004 @@
...
+    workingArray[positionX] = 2;
+  } else {
+    page.querySelectorAll('.field-cell')[positionX]
+      .classList.add('field-cell--4');
+    workingArray[positionX] = 4;
+  }
+};
+
+function removeAllModificators() {
+  workingArray = Array(16).fill(null);
```
lerastarynets, line 79: why are you reassigning lets in almost every function if you already declared the same value?

src/scripts/main.js
```diff
@@ -1,3 +1,1004 @@
...
+
+      page.querySelectorAll('.field-cell')[currentCellIndex]
+        .classList.remove(currentModifier);
+      workingArray[nearestCellIndex] *= 2;
+      workingArray[currentCellIndex] = null;
+      changeIndicator = true;
+      scoreCounter += workingArray[nearestCellIndex];
+      switcherScore.textContent = scoreCounter;
+
+      if (workingArray[indexThroughOneCell] === null) {
```
lerastarynets, line 371: Same here and for all such cases with `null` as it's a falsy value
```suggestion
      if (!workingArray[indexThroughOneCell]) {
```

lerastarynets (changes requested):
Nice start) For now your `main.js` file has 1000+ lines and is barely understandable which is bad practice. Pls consider reducing size of the file. For example you can declare  field like so
 ```
gameField = [
    [0, 0, 0, 0],
    [0, 0, 0, 0],
    [0, 0, 0, 0],
    [0, 0, 0, 0],
  ];
```
and then manipulate its' rows and columns. Also you can declare general `move` function and execute it with `switch case` depending on `direction` passed as an argument to the func

File: src/scripts/main.js (hunk 6 above)
Comment (reseneweb) [reply src/scripts/main.js:79]: Valeria, hello! I reassign the value of workingArray = Array(16).fill(null) to the same as I declared at the beginning in order to return (update) the initial value of workingArray when Restart is clicked. Otherwise, the array will store the old data. I just reset workingArray to its initial position to restart the game.

File: src/scripts/main.js (hunk 7 above)
Comment (reseneweb) [reply src/scripts/main.js:371]: Hello! I tried to substitute the value (!workingArray[indexThroughOneCell]) instead of (workingArray[indexThroughOneCell] === null). By analogy, I did it everywhere. Unfortunately, the program has completely stopped working.

The matter is that check on 'null' is necessary to me. I am filling all empty cells of the working field with null values. All values outside of the working array are undefined. For the program to work correctly, I'm looking for exactly the null value. If (workingArray[indexThroughOneCell] === null) then we move the cell and sum the value. I deliberately fill empty array cells with 'null' values to distinguish them from external values - undefined. Therefore, I'm looking for exactly null, and if the value is null, then we move the cell inside the square. External values will not be affected because they have a value - undefined.

Снимок

File: src/scripts/main.js [hunk 8]
```diff
@@ -1,3 +1,466 @@
...
+
+        indexThroughTwoCells = allIndices[i] - 3;
+
+        if ((currentCellIndex === 0) || (currentCellIndex === 1)
+        || (currentCellIndex === 2) || (currentCellIndex === 4)
+        || (currentCellIndex === 5) || (currentCellIndex === 6)
+        || (currentCellIndex === 8) || (currentCellIndex === 9)
+        || (currentCellIndex === 10) || (currentCellIndex === 12)
+        || (currentCellIndex === 13) || (currentCellIndex === 14)
+        ) {
```
sTorba24, line 225: probably easier to check if value of cell isn't equal 3, 7, 11 etc..

sTorba24 (approved):
I approve it 

but solution is too complicated  to read ;(
Would be good to rework it for portfolio

PR #398, opened 2023-07-05
2048 js game

sofiatryasko:
[DEMO LINK](https://sofiatryasko.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,190 @@
 'use strict';
 
-// write your code here
+const field = document.querySelector('.game-field').firstElementChild;
+const cellList = document.querySelectorAll('TD');
+const startButton = document.querySelector('.start');
+const score = document.querySelector('.game-score');
+const rows = [...field.rows].reduce((res, el, index) => {
+  if (!res[index]) {
+    res[index] = [];
+  };
+
+  [...el.children].forEach((item) => {
+    res[index].push(item);
+  });
+
+  return res;
+}, []);
+const cols = [...field.rows].reduce((res, el) => {
+  [...el.children].forEach((item, index) => {
+    if (!res[index]) {
+      res[index] = [];
+    };
+
+    res[index].push(item);
+  });
+
+  return res;
+}, []);
+
+startButton.addEventListener('click', e => {
+  if (!startButton.toggle) {
+    document.addEventListener('keydown', game);
+    showMessage();
+    start();
+    startButton.innerText = 'Restart';
+    startButton.classList.remove('start');
+    startButton.classList.add('restart');
+    startButton.toggle = !startButton.toggle;
+  } else {
+    document.removeEventListener('keydown', game);
+    document.addEventListener('keydown', game);
+    score.innerText = 0;
+    cellList.forEach(el => (el.innerText = ''));
+    start();
+    showMessage();
+  }
+});
+
+function game(e) {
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,190 @@
 'use strict';
 
-// write your code here
+const field = document.querySelector('.game-field').firstElementChild;
+const cellList = document.querySelectorAll('TD');
+const startButton = document.querySelector('.start');
+const score = document.querySelector('.game-score');
+const rows = [...field.rows].reduce((res, el, index) => {
```
VitaliyBodnar, line 7: try to use a more descriptive name than 'el'
in this case you can use 'row'
please check other places

src/scripts/main.js
```diff
@@ -1,3 +1,190 @@
...
+        clear[i] = +clear[i] + +clear[i - 1];
+        clear[i - 1] = '';
+      };
+    }
+
+    clear = clear.filter(item => {
+      if (item) {
+        return item;
+      }
+    });
```
VitaliyBodnar, line 77:
```suggestion
    clear = clear.filter(item => item);
```

demetrxx (changes requested):
Seems like bug. 

image

src/scripts/main.js
```diff
@@ -1,3 +1,190 @@
...
+            : el[ind].innerText = '';
+        }
+        break;
+    }
+  });
+
+  cellClassHandler();
+
+  cellList.forEach(el => {
+    if (+el.innerText === 2048) {
```
VitaliyBodnar, line 102: '2048' is a 'magic number', so please move it to the constant

src/scripts/main.js
```diff
@@ -1,3 +1,190 @@
...
+function addNewNumber(array) {
+  const emptyCells = findEmptyCell(array);
+  const position = randomCell(emptyCells.length);
+  let number = randomCell(10);
+
+  if (number <= 1) {
+    number = 4;
+  } else {
+    number = 2;
+  }
```
VitaliyBodnar, line 135:
```suggestion
  number = number <= 1 ? 4 : 2
```
- please also put the '4' and '2' to the const, because it will make the code cleaner

VitaliyBodnar (changes requested):
Well done!

Anastasiia-Svintsova (changes requested):
- Hm, there are available movements, but the loss message appeared 🤔 
image

- Disable clicks on buttons other than arrows. Here there is an issue when pressing random keys on the keyboard, watch the video below
chrome-capture-2023-6-5 (2)

sTorba24 (changes requested):
Good work! Lets improve it

now new item appears just on `click` 

but it should be added if there are items to connect or place to move them
try to reproduce it clicking same direction button(right, for example) a lot of times
and then check original game
https://2048game.com/

anastasiia-tilikina (commented):
the issues from previous review are not fixed

anastasiia-tilikina (changes requested):
.

VitaliyBondarenko1982 (changes requested):
Looks like still not fixed prev comment or deploy not updated.
After click on start button should appear two cells with numbers.
But they appears after move or click.
chrome-capture-2023-10-1

Esceype (approved):
Well done!

PR #403, opened 2023-07-09
Write_down_solution_for_2048_game

rialleons:
- [DEMO LINK](https://rialleons.github.io/js_2048_game/)


VitaliyBondarenko1982 (changes requested):
Good job.
Check such case
When after press on some arrow cells no move - shouldn't exise new numbers.
On screencase i pressed on top on the start of video
chrome-capture-2023-6-10

IvanFesenko (approved):

PR #407, opened 2023-07-11
Develop

DenisGurskiy:
https://DenisGurskiy.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,8 +1,6 @@
 import { createBox } from './box';
 import { handleInput } from './move';
 
-const GRID = 4;
-const COUNT_GRID = GRID * GRID;
 let matrix = [
   [0, 0, 0, 0],
   [0, 0, 0, 0],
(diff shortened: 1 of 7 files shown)


src/styles/main.scss
```diff
@@ -10,17 +13,59 @@ body {
...
   background: #d6cdc4;
-  width: 75px;
-  height: 75px;
+  width: $cell-size;
+  height: $cell-size;
   border-radius: 5px;
   color: #776e65;
   box-sizing: border-box;
   text-align: center;
   vertical-align: center;
```
volodymyr-soltys97, line 34: vertical-align: center is not a valid value. It should be middle instead of center.

src/styles/main.scss
```diff
@@ -10,17 +13,59 @@ body {
   font-weight: 900;
 }
 
+.game-field {
+  background: #bbada0;
+  border-spacing: 10px;
```
volodymyr-soltys97, line 18: border-spacing style is used but it will have no effect because it applies only to table elements. You probably want to use margin or padding.

src/styles/main.scss
```diff
@@ -10,17 +13,59 @@ body {
...
+    }
+  }
+}
+
+.field-tile {
+  position: absolute;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background: yellow;
```
volodymyr-soltys97, line 50: The background color is set to yellow, but it is not clear why this color has been chosen. Please add a comment to explain or consider using a variable for colors.

src/scripts/main.js
```diff
@@ -1,3 +1,188 @@
...
+    });
+
+    return emptyCoordinates;
+  }, []);
+}
+
+export function setBoxToMatrix(box) {
+  const x = box.position.x ? box.position.x : 0;
+  const y = box.position.y ? box.position.y : 0;
+  // const { x, y } = box.position;
```
volodymyr-soltys97, line 89:
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

volodymyr-soltys97 (approved):
Nice work 🔥

PR #413, opened 2023-07-19
2048

w315:
[DEMO LINK](https://w315.github.io/js_2048_game/)


VitaliyBondarenko1982 (changes requested):
Good job.
Found one bug.
If no cells for merge and not possible to move any cell - shouldn't appear new number.
Look to screencast. I pressed top arrow and appear in random empty cell  `2, but shouldn't
chrome-capture-2023-6-19 (1)

volodymyr-soltys97 (approved):
Nice work 🔥

PR #414, opened 2023-07-20
first commit

viktoria-krasnyuk:
https://viktoria-krasnyuk.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,282 @@
 'use strict';
 
-// write your code here
+const gameField = document.querySelector('.game-field');
+const scoreGame = document.querySelector('.game-score');
+const button = document.querySelector('.button');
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,282 @@
...
+  if (e.code === 'ArrowLeft') {
+    slideLeft();
+    setTwoOrFour();
+  } else if (e.code === 'ArrowRight') {
+    slideRight();
+    setTwoOrFour();
+  } else if (e.code === 'ArrowUp') {
+    slideUp();
+    setTwoOrFour();
+  } else if (e.code === 'ArrowDown') {
```
polosanya, line 155: It is better to use switch...case here instead as you have a limited amount of codes

polosanya (approved):
Great!

PR #420, opened 2023-07-26
add task solution

Luk2asz:
- [DEMO LINK](https://Luk2asz.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,230 @@
 'use strict';
 
-// write your code here
+const button = document.querySelector('.button');
+const message = document.querySelector('.message-start');
+const lose = document.querySelector('.message-lose');
+const win = document.querySelector('.message-win');
+const rows = document.querySelectorAll('tr');
+const allCells = document.querySelectorAll('td');
+let buttonClicked = false;
+
+button.addEventListener('click', () => {
+  buttonClicked = true;
+  message.classList.add('hidden');
+
+  if (!buttonClicked) {
+    newCell();
+    newCell();
+  } else {
+    reset();
+  }
+});
+
+function isGameOver() {
+  let gameOver = true;
+
+  for (let i = 0; i < 4; i++) {
+    for (let j = 0; j < 4; j++) {
+      if (!rows[i].children[j].textContent) {
+        gameOver = false;
+        break;
+      }
+    }
+  }
+
+  if (gameOver === true) {
+    for (let i = 0; i < 4; i++) {
+      for (let j = 0; j < 3; j++) {
+        const checkCell = rows[i].children[j].textContent;
+        const nextCell = rows[i].children[j + 1].textContent;
+
+        if (checkCell === nextCell) {
+          gameOver = false;
+          break;
+        }
+      }
+    }
+  }
+
+  if (gameOver === true) {
+    for (let i = 0; i < 3; i++) {
+      for (let j = 0; j < 4; j++) {
+        const checkCell = rows[i].children[j].textContent;
+        const nextCell = rows[i + 1].children[j].textContent;
+
+        if (checkCell === nextCell) {
+          gameOver = false;
+          break;
+        }
+      }
+    }
+  }
+
+  if (gameOver === true) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+    rows[xPosition].cells[yPosition].textContent = TwoOrFour();
+    classForCell();
+    isGameOver();
+  } else {
+    newCell();
+  }
+}
+
+function TwoOrFour() {
+  if (Math.random() <= 0.1) {
```
darokrk, line 82: The function TwoOrFour() doesn't follow JavaScript naming conventions. Consider renaming it to twoOrFour() to follow camelCase convention.

src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+          gameOver = false;
+          break;
+        }
+      }
+    }
+  }
+
+  if (gameOver === true) {
+    lose.classList.remove('hidden');
+  }
```
darokrk, line 65: The isGameOver() function only shows a game over message but doesn't actually stop the game. Consider making this function return a boolean and use that to stop the game when true.

src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
-// write your code here
+const button = document.querySelector('.button');
+const message = document.querySelector('.message-start');
+const lose = document.querySelector('.message-lose');
+const win = document.querySelector('.message-win');
+const rows = document.querySelectorAll('tr');
+const allCells = document.querySelectorAll('td');
+let buttonClicked = false;
+
+button.addEventListener('click', () => {
```
darokrk, line 11: buttonClicked is set to true at the start of the function, so the if (!buttonClicked) condition will never be true. You should review the logic here.

src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+
+function TwoOrFour() {
+  if (Math.random() <= 0.1) {
+    return 4;
+  } else {
+    return 2;
+  }
+}
+
+function reset() {
```
darokrk, line 89: The scoreValue variable is not defined in this file. Ensure that it is defined globally or passed in as a parameter.

src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+const rows = document.querySelectorAll('tr');
+const allCells = document.querySelectorAll('td');
+let buttonClicked = false;
+
+button.addEventListener('click', () => {
+  buttonClicked = true;
+  message.classList.add('hidden');
+
+  if (!buttonClicked) {
+    newCell();
```
darokrk, line 16: do we need to call this function twice?

src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+  scoreValue = 0;
+  score.textContent = scoreValue;
+  win.classList.add('hidden');
+  lose.classList.add('hidden');
+
+  allCells.forEach((cell) => {
+    cell.textContent = '';
+  });
+  newCell();
+  newCell();
```
darokrk, line 99: same here

src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+
+        if (checkCell === nextCell) {
+          gameOver = false;
+          break;
+        }
+      }
+    }
+  }
+
+  if (gameOver === true) {
```
darokrk, line 49: You defined three of the same conditional checks `if (gameOver === true)`, maybe we should use the all code in just one if if they required same condition ?

darokrk (changes requested):
The game works 👍🏽  but code need few more improvements 🛠️ 👨🏽‍💻

src/scripts/main.js
Luk2asz: at end of all conditional checks are code: 
`if (checkCell === nextCell) {
          gameOver = false;
          break;
        }`
if this condition will true, other conditional checks will not apply. Otherwise all three conditional checks will senseles apply.

darokrk (approved):
Good work 🏆

PR #422, opened 2023-07-28
2048 the game

misha-shevchuk:
[DEMO LINK](https://misha-shevchuk.github.io/js_2048_game/)

----------------------------------------------------------------
I've decided to use "keyup" event instead of "keydown" just in case to avoid continuous moves if press and hold any arrow key. Also, I`ve changed UI a little and added some sound effects for a better experience. 

p.s. During development I have never won the game, maybe you will have more luck ;)


src/scripts/main.js
```diff
@@ -1,3 +1,210 @@
...
+
+function getRandomCell() {
+  return clearCells[Math.floor(Math.random() * clearCells.length)];
+}
+
+function getCellValue() {
+  return Math.random() < 0.9 ? 2 : 4;
+}
+
+function newCell() {
```
volodymyr-soltys97, line 33: The condition 'clearCells.length === undefined' can be simplified to '!clearCells.length' which checks if the array is empty or not.

src/scripts/main.js
```diff
@@ -1,3 +1,210 @@
...
+      }
+      break;
+  }
+
+  cells.forEach(cell => {
+    cell.removeAttribute('data-blocked');
+  });
+}
+
+document.addEventListener('keyup', (evnt) => {
```
volodymyr-soltys97, line 191: It's more efficient to listen to the keyup event only on the window or document object, not on the entire document. So you can change 'document.addEventListener' to 'window.addEventListener' or 'document.body.addEventListener'.

src/scripts/main.js
```diff
@@ -1,3 +1,210 @@
...
+    startGame();
+  } else {
+    restartGame();
+  }
+});
+
+function canBeMarged() {
+  const wholeField = [...cellsInRow, ...columns];
+
+  for (const line of wholeField) {
```
volodymyr-soltys97, line 85: Avoid using a `for...of` loop to iterate over an array. Instead, use `forEach` or `map`.

src/scripts/main.js
```diff
@@ -1,3 +1,210 @@
...
+function getDirection(direction) {
+  if (!clearCells.length && !canBeMarged()) {
+    loseMessage.classList.remove('hidden');
+    looseSound.currentTime = 0;
+    looseSound.play();
+  }
+
+  switch (direction) {
+    case 'ArrowUp':
+      for (const column of columns) {
```
volodymyr-soltys97, line 162: Avoid using a `for...of` loop to iterate over an array. Instead, use `forEach` or `map`.

src/scripts/main.js
```diff
@@ -1,3 +1,210 @@
...
+
+function getDirection(direction) {
+  if (!clearCells.length && !canBeMarged()) {
+    loseMessage.classList.remove('hidden');
+    looseSound.currentTime = 0;
+    looseSound.play();
+  }
+
+  switch (direction) {
+    case 'ArrowUp':
```
volodymyr-soltys97, line 161: Сreate a variable for these magic words

volodymyr-soltys97 (changes requested):
Nice work 👍 
Let's improve your code

src/scripts/main.js
misha-shevchuk: I've decided to change it back to `'keydown'`. It feels more natural and more responsive while playing.

src/scripts/main.js
```diff
@@ -1,3 +1,215 @@
...
+      cellsInRow.forEach((row) => {
+        flipLine(row);
+      });
+      break;
+
+    case left:
+      cellsInRow.forEach((row) => {
+        flipLine([...row].reverse());
+      });
+      break;
```
polosanya, line 188: Switch case should always have default case to handle possible errors

polosanya (changes requested):
It is possible to play after the loss. Make sure that you handle the losing correctly
P.S. Sounds feature is crazy 🔥 Well done!
loss

volodymyr-soltys97 (approved):
Nice work 🔥

volodymyr-soltys97 (approved):
Nice work 🔥

PR #424, opened 2023-07-29
Solution 29072023

Dmytro-Budynskyi:


4ksandrey (changes requested):
npm test has failed, check it, fix and then send task on review

PR #426, opened 2023-07-30
add task solution

RomanenkoStud:
[DEMO LINK](https://RomanenkoStud.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/Field.js
+++ b/src/scripts/Field.js
@@ -1,0 +1,155 @@
+'use strict';
+
+const Cell = require('./Cell');
+
+class Field {
+  constructor(element) {
+    this.cells = [];
+
+    const rows = element.querySelectorAll('.field-row');
+
+    for (let i = 0; i < rows.length; i++) {
+      const row = rows[i].querySelectorAll('.field-cell');
+
+      this.cells.push([...row].map((cell) => new Cell(cell)));
+    }
+  }
+
+  getMaxValue() {
+    let maxValue = 0;
+
+    for (let row = 0; row < this.cells.length; row++) {
+      for (let column = 0; column < this.cells[row].length; column++) {
+        const cellValue = this.cells[row][column].value;
+
+        if (cellValue > maxValue) {
+          maxValue = cellValue;
+        }
+      }
+    }
+
+    return maxValue;
+  }
+
+  hasAvailableMoves() {
+    for (let row = 0; row < this.cells.length; row++) {
+      for (let column = 0; column < this.cells[row].length; column++) {
+        const cell = this.cells[row][column];
+
+        if (cell.isEmpty) {
+          return true;
+        }
+
+        if (
+          (row > 0 && this.cells[row - 1][column].value === cell.value)
+          || (column > 0 && this.cells[row][column - 1].value === cell.value)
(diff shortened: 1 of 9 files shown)


polosanya (changes requested):
Nice start! Let's improve the game a bit
1. Game doesn't work correctly. Numbers are not combined if they should move before.
Check how it works on working example
2048-bug
2. I can press any button and add a number
image
3. There should be a chance for "4" to appear at start

src/scripts/Game.js
```diff
@@ -0,0 +1,112 @@
...
+    this.button = new Button(
+      document.querySelector('.controls .button'),
+      onStart, onRestart
+    );
+
+    this.controls = new SwipeControls(onLeft, onRight, onUp, onDown, onAction);
+    this.message.setMessage(this.MESSAGES.START);
+  }
+
+  load() {
```
volodymyr-soltys97, line 107: The load() method returns this without doing anything, which seems unnecessary. If this is not being used, consider removing it.

src/scripts/Field.js
```diff
@@ -0,0 +1,154 @@
...
+      totalScore += this.mergeCells(columnCells);
+    }
+
+    return totalScore;
+  }
+
+  addTile(value) {
+    const emptyCells = [];
+
+    for (let row = 0; row < this.cells.length; row++) {
```
volodymyr-soltys97, line 123: Consider using 'forEach' or 'map' instead of the 'for' loop for better readability and performance. You could use a nested 'forEach' loop to iterate over the 2D 'cells' array.

src/scripts/Field.js
```diff
@@ -0,0 +1,154 @@
...
+          maxValue = cellValue;
+        }
+      }
+    }
+
+    return maxValue;
+  }
+
+  hasAvailableMoves() {
+    for (let row = 0; row < this.cells.length; row++) {
```
volodymyr-soltys97, line 35: Consider using 'forEach' or 'map' instead of the 'for' loop for better readability and performance. You could use a nested 'forEach' loop to iterate over the 2D 'cells' array.

src/scripts/Field.js
```diff
@@ -0,0 +1,154 @@
...
+
+const Cell = require('./Cell');
+
+class Field {
+  constructor(element) {
+    this.cells = [];
+
+    const rows = element.querySelectorAll('.field-row');
+
+    for (let i = 0; i < rows.length; i++) {
```
volodymyr-soltys97, line 11: Consider using 'forEach' or 'map' instead of the 'for' loop for better readability and performance. For instance, you can refactor this part like 'rows.forEach((row, i) => {...})'.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

polosanya (changes requested):
1. Sometimes random number appears on move. Please make sure that only 2 or 4 appears
2048-random
2. Also, "4" still never appears at start
image

RomanenkoStud:
sorry, i thought for some reason that any number can appear)

volodymyr-soltys97 (approved):
Good job 👍

PR #435, opened 2023-08-04
solution

danulyk05:
https://danulyk05.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -43,22 +43,18 @@ document.addEventListener('keydown', (e) => {
   switch (e.key) {
     case 'ArrowLeft':
       movedLeft();
-      setRandomTwo();
       break;
 
     case 'ArrowRight':
       movedRight();
-      setRandomTwo();
       break;
 
     case 'ArrowUp':
       movedUp();
-      setRandomTwo();
       break;
 
     case 'ArrowDown':
       movedDown();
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,246 @@
...
+      field[r][c] = 2;
+      renderField();
+
+      found = true;
+    }
+  }
+};
+
+function filterZero(row) {
+  return row.filter(num => num !== 0);
```
loralevitska, line 110:
```suggestion
  return row.filter(num => num);
```
OR
```suggestion
  return row.filter(num => !!num);
```

src/scripts/main.js
```diff
@@ -1,3 +1,246 @@
...
+        return false;
+      }
+
+      if (col > 0 && board[row][col - 1] === currentCell) {
+        return false;
+      }
+
+      if (col < size - 1 && board[row][col + 1] === currentCell) {
+        return false;
+      }
```
loralevitska, line 241: Combine similar cases into 1, if condition will be so long use helper variables with meaningful variable naming 
image

loralevitska (changes requested):
- remove outline, it breaks design  
image

etojeDenys (changes requested):
if the elements have not moved, a new one should not be added
image

PR #436, opened 2023-08-04
Develop

RuslanMelnykFE:


nazarmatsevych (changes requested):
Code looks interesting but demo link doesn't work, try to fix it and put this link to your pull request description

src/scripts/main.js
```diff
@@ -1,3 +1,472 @@
 'use strict';
 
-// write your code here
+//==============================================================
```
TarasHoliuk, line 3: use regions on development stage, but delete it before pushing changes

src/scripts/main.js
```diff
@@ -1,3 +1,472 @@
...
+
+    this.viewKeyEvents.keyEvent();
+  }
+
+  init() {};
+}
+
+const view = new View();
+
+view.init();
```
TarasHoliuk, line 469: it seems like you call an empty method

TarasHoliuk (changes requested):
your demo link still doesn't work - feel free to ask in the chat if you need help
image

PR #439, opened 2023-08-10
Develop

Hladkyi03:
[DEMO LINK](https://Hladkyi03.github.io/js_2048_game/)


maxim2310 (changes requested):
May be you forgot tot deploy last version
image

Hladkyi03:
So sorry, it was bug with gh-pages my formulas for top and left styles wasn`t the same that in scss  code, so i applied it in inline styles through JS. Mentor said it okay, bcus it isn`t problem with code but with gh-pages.
If you want you can check the thread: https://mate-parttime.slack.com/archives/C03UQS59NRY/p1691665987563989

etojeDenys (approved):
lgtm

PR #451, opened 2023-08-21
solution

tYuriy:


etojeDenys (changes requested):
try to re-deploy your site cause link does not work

sTorba24 (changes requested):
Well done! Lets make it perfect

Now new cells appear with every click on button (try press only 'right', for example)
But it should has a restriction - if existing cells hasn't place to move and nothing to connect - it shouldn't add new one

Check how original game works https://2048game.com/

TarasHoliuk (approved):
:fire:

PR #452, opened 2023-08-23
add task

VladyslavKolisnyk:
DEMO LIMK: https://VladyslavKolisnyk.github.io/js_2048_game/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,223 @@
 'use strict';
 
-// write your code here
+const button = document.querySelector('button');
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const gameScore = document.querySelector('.game-score');
+const fieldRows = document.querySelectorAll('.field-row');
+const size = 4;
+let score = 0;
+let isWin = false;
+let newGameField;
+let gameField = [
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,223 @@
...
+        elem.className = `field-cell field-cell--${cell}`;
+      }
+    };
+  };
+
+  gameScore.textContent = score;
+};
+
+function left() {
+  if (!checkRows()) {
```
volodymyr-soltys97, line 140: The function 'left' is not clearly defined. It is better to use more descriptive function names.

src/scripts/main.js
```diff
@@ -1,3 +1,223 @@
...
+  addNumber();
+  addNumber();
+  render();
+});
+
+function move(e) {
+  newGameField = gameField;
+
+  switch (e.key) {
+    case 'ArrowLeft':
```
volodymyr-soltys97, line 41: Create a variable for these magic words and use them everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #455, opened 2023-08-24
solution

gabriella135:
[DEMO LINK](https://gabriella135.github.io/js_2048_game/)


maxim2310 (changes requested):
Good job, but would be nice to add hint text in case you cant move anymore 
image

lerastarynets (approved):
You're doing a fantastic job, keep going!

PR #461, opened 2023-08-28
2048

Duster-1:
https://duster-1.github.io/js_2048_game/


maxim2310 (approved):
Great job overall! Just one tiny thing needs to be fixed)
image
make these buttons brighter

PR #464, opened 2023-08-31
Did I do this?

YuraMisura:
- [DEMO LINK](https://YuraMisura.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -53,6 +53,7 @@ function goStart() {
 function goEnd() {
   setButtonStatus('Start');
   message[0].classList.remove('hidden');
+  message[2].classList.add('hidden');
   matrix = matrix.map(row => row.map(() => 0));
   resetGameFieldCells();
   gameScore = 0;
@@ -120,6 +121,11 @@ function upgradeFeilds() {
 }
 
 document.addEventListener('keydown', click => {
+  if (buttonStart.classList.contains('game-start')) {
+    setButtonStatus('Restart');
+    message[0].classList.add('hidden');
+  }
+
   if (!areMovesAvailable()) {
     message[2].classList.remove('hidden');
     message[0].classList.add('hidden');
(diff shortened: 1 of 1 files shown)


YuraMisura:
> Hello, to be fair, I'm exhausted. Could you look at this code and if I'm doing this right, say 'carry on,' or 'you should do it like this,' or 'try this way'?

- [DEMO LINK](https://YuraMisura.github.io/js_2048_game/)

src/scripts/main.js
```diff
@@ -1,3 +1,163 @@
...
+      row.push(0);
+    }
+  });
+
+  matrix = result;
+  upgradeFeilds();
+  addRandomCell();
+}
+
+function moveRight() {
```
IvanFesenko, line 153: The functions 'moveRight', 'moveUp', and 'moveDown' are not yet implemented. Implement these functions to enable the corresponding movements in the game.

src/scripts/main.js
```diff
@@ -1,3 +1,163 @@
...
+  matrix = matrix.map(row => row.map(() => 0));
+
+  gameFieldCells.forEach(cell => {
+    cell.textContent = '';
+    cell.className = 'field-cell';
+  });
+  gameScore = 0;
+  addRandomCell();
+  addRandomCell();
+};
```
IvanFesenko, line 57: You're adding random cells after resetting the game, but you haven't considered the case when a player may lose right after the game restarts. You should check if there are valid moves before adding new random cells.

src/scripts/main.js
```diff
@@ -1,3 +1,163 @@
...
+buttonStart.addEventListener('click', () => {
+  if (buttonStart.classList.contains('start')) {
+    goStart();
+  } else {
+    goEnd();
+  }
+});
+
+function goStart() {
+  buttonStart.classList.remove('start');
```
IvanFesenko, line 33: It is better to use a separate class for game status and toggle it instead of manually adding and removing classes 'start' and 'restart'.

src/scripts/main.js
```diff
@@ -1,3 +1,163 @@
...
+  buttonStart.classList.add('restart');
+  buttonStart.textContent = 'restart';
+
+  message[0].classList.add('hidden');
+  addRandomCell();
+};
+
+function goEnd() {
```
IvanFesenko, line 41: Same issue as the previous one. Use a separate class for game status and toggle it instead of manually adding and removing classes 'start' and 'restart'.

src/scripts/main.js
```diff
@@ -1,3 +1,163 @@
...
+const buttonStart = document.querySelector('.start');
+const gameFieldCells = document.querySelectorAll('.field-cell');
+const infoScore = document.querySelector('.game-score');
+let gameScore = +(infoScore.textContent);
+const ROW_LENGTH = 4;
+const PROBILITY_NUMBER = 0.1;
+
+addRandomCell();
+addRandomCell();
+
```
IvanFesenko, line 23: Create new cell when game starts

IvanFesenko (changes requested):
Keep going, do not give up 😉 
Pay attention on comment also:
- cell updated before click start

YuraMisura:
I`ve made some alteresions and add some functions. 
I am having trouble with these cases:  "The move is possible if at least one cell is changed after the move"   and "The merged cell can’t be merged twice during one move".
P.S. sorry for my English

VitaliyBondarenko1982 (changes requested):
Before add random cell after upgrade fields you can compare each cell in prev matrix with actual matrix. if some cell not equal with prev - you need to add random cell. if all cells are equals - it means that wasn't any move.

maxim2310 (changes requested):
Good work!!! but have some comments
1. you can start game before click on start button
2. you must to move all cell without spaces when some cells are combined
image
3. when clicked on restart you must hide this text blocks
image

etojeDenys (approved):
good job

PR #465, opened 2023-09-03
add task solution

MishaLehotskyi:


src/index.html
```diff
@@ -12,50 +12,50 @@
...
       </div>
 
       <table class="game-field">
         <tbody>
           <tr class="field-row">
-            <td class="field-cell"></td>
-            <td class="field-cell"></td>
-            <td class="field-cell"></td>
-            <td class="field-cell"></td>
+            <td id="0-0" class="field-cell"></td>
```
mykhalenych, line 24: Adding ids for individual table cells may not be the most efficient way to manage cell data. Consider handling this in JavaScript with an array or object instead.

src/index.html
```diff
@@ -12,50 +12,50 @@
         <h1>2048</h1>
         <div class="controls">
           <p class="info">
-            Score: <span class="game-score">0</span>
+            Score: <span id="game-score" class="game-score">0</span>
           </p>
-          <button class="button start">Start</button>
+          <button id="button-start" class="button start">Start</button>
```
mykhalenych, line 17: the same

src/index.html
mykhalenych: Screenshot 2023-09-04 at 16 31 34

you can use .classList.add(`field-cell--${num}`) and .classList.replace

mykhalenych: you already do this way, delete all ids

MishaLehotskyi: cannot delete id's because it is main part my program's logic

maxim2310 (approved):
Good work, but next time add demolink to PR)

PR #472, opened 2023-09-10
Develop

GeorgeSavostikov:
[DEMO LINK](https://GeorgeSavostikov.github.io/js_2048_game/)


etojeDenys (changes requested):
Screenshot 2023-09-10 at 11 10 27
if we don't have any element to merge, we shouldn't create a new cell

IvanFesenko (changes requested):
So when is victory will be?
https://github.com/mate-academy/js_2048_game/assets/61904995/c5644599-f7ad-41e1-b8ed-80d254366855

GeorgeSavostikov:
image

To be precise, the condition says that the winning message is displayed when at least one cell in the field has the value 2048.

lerastarynets (approved):
lgtm 🚀

PR #476, opened 2023-09-13
Solution

NataliiaHen:
[DEMO LINK](https://NataliiaHen.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,359 @@
 'use strict';
 
-// write your code here
+let board = [];
+const boardSize = 4;
+let score = 0;
+let gameOver = false;
+let gameStarted = false;
+
+const startButton = document.querySelector('.start');
+const restartButton = document.querySelector('.restart');
+const scoreEl = document.querySelector('.game-score');
+
+const startMessage = document.querySelector('.message.message-start');
+const winMessage = document.querySelector('.message-win');
+const loseMessage = document.querySelector('.message-lose');
+
+const gameBoard = document.querySelector('.game-field');
+const rows = gameBoard.querySelectorAll('.field-row');
+
+window.addEventListener('load', () => {
+  initializedBoard();
+  generateTwoTile();
+  upDateBoard();
+});
+
+document.addEventListener('keydown', changeKey);
+
+startButton.addEventListener('click', () => {
+  startGame();
+}
+);
+
+restartButton.addEventListener('click', () => {
+  restartGame();
+});
+
+function startGame() {
+  toggleButton(startButton, restartButton);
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,359 @@
...
+        moveDown();
+        break;
+      case 'ArrowLeft':
+        moveLeft();
+        break;
+      case 'ArrowRight':
+        moveRight();
+        break;
+    }
+    generateTwoTile();
```
etojeDenys, line 82: only one cell should appear when user press some button

etojeDenys (changes requested):
if we don't have any cells to merge after move then we don't need to add 

https://github.com/mate-academy/js_2048_game/assets/70220667/4c14e3aa-cb6d-43b5-9093-c3088304dc62

another cell

src/scripts/main.js
NataliiaHen: The game start with one cell?

src/scripts/main.js
```diff
@@ -1,3 +1,339 @@
...
+const rows = gameBoard.querySelectorAll('.field-row');
+
+document.addEventListener('keydown', changeKey);
+
+startButton.addEventListener('click', () => {
+  toggleButton(startButton, restartButton);
+  hideElem(startMessage);
+  startGame();
+}
+);
```
loralevitska, line 27:
```suggestion
});
```

src/scripts/main.js
```diff
@@ -1,3 +1,339 @@
...
+        break;
+      case 'ArrowDown':
+        moveDown();
+        break;
+      case 'ArrowLeft':
+        moveLeft();
+        break;
+      case 'ArrowRight':
+        moveRight();
+        break;
```
loralevitska, line 70: Would be great to add default case too
```suggestion
      case 'ArrowUp':
        moveUp();
        break;
        
      case 'ArrowDown':
        moveDown();
        break;
        
      case 'ArrowLeft':
        moveLeft();
        break;
        
      case 'ArrowRight':
        moveRight();
        break;
```

src/scripts/main.js
```diff
@@ -1,3 +1,339 @@
...
+      merged = false;
+
+      if (currentCell !== 0) {
+        let currentRow = row;
+
+        while (
+          currentRow > 0
+          && (board[row - 1][col] === 0
+            || (board[row - 1][col] === board[row][col]
+              && !merged))
```
loralevitska, line 105: Long expression move to variable. Fix all cases

loralevitska (approved):
Good job!

PR #478, opened 2023-09-15
2048 game

AnastasiiaKriuchkova:
[DEMO LINK](https://AnastasiiaKriuchkova.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,248 @@
 'use strict';
 
-// write your code here
+const fieldRows = document.querySelectorAll('.field_row');
+const messageWin = document.querySelector('.message_win');
+const messageLose = document.querySelector('.message_lose');
+const button = document.querySelector('.start');
+const messageStart = document.querySelector('.message_start');
+const gameScore = document.querySelector('.game_score');
+
+const GRID_SIZE = 4;
+
+document.addEventListener('keydown', (e) => {
+  if (e.key === 'ArrowUp') {
+    grid.slideUp();
+  } else if (e.key === 'ArrowLeft') {
+    grid.slideLeft();
+  } else if (e.key === 'ArrowDown') {
+    grid.slideDown();
+  } else if (e.key === 'ArrowRight') {
+    grid.slideRight();
+  }
+});
+
+button.addEventListener('click', () => {
+  button.classList.replace('start', 'restart');
+  button.textContent = 'Restart';
+  messageStart.classList.add('hidden');
+  messageWin.classList.add('hidden');
+  messageLose.classList.add('hidden');
+  grid.startGame();
+});
+
+class Grid2048 {
+  startGame() {
+    this.field = [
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,248 @@
...
+const button = document.querySelector('.start');
+const messageStart = document.querySelector('.message_start');
+const gameScore = document.querySelector('.game_score');
+
+const GRID_SIZE = 4;
+
+document.addEventListener('keydown', (e) => {
+  if (e.key === 'ArrowUp') {
+    grid.slideUp();
+  } else if (e.key === 'ArrowLeft') {
```
OleziO, line 15: Use switch case construction. Move all cases to the object.

Example of usage:

`case Directions.ArrowLeft:`

OleziO (changes requested):
Fix all linter errors.

AnastasiiaKriuchkova:
> Fix all linter errors.

I fixed the linter errors, but it appears another one.  i have no idea how to solve this

src/scripts/main.js
```diff
@@ -1,3 +1,253 @@
...
+          return false;
+        }
+
+        if (rowIndex !== GRID_SIZE - 1) {
+          if (
+            fieldCopy[rowIndex][colIndex] === fieldCopy[rowIndex + 1][colIndex]
+          ) {
+            return false;
+          }
+        }
```
DarkMistyRoom, line 115: looks like you can combine these conditions

src/scripts/main.js
```diff
@@ -1,3 +1,253 @@
...
+
+        if (colIndex !== GRID_SIZE - 1) {
+          if (
+            fieldCopy[rowIndex][colIndex] === fieldCopy[rowIndex][colIndex + 1]
+          ) {
+            return false;
+          }
+        }
```
DarkMistyRoom, line 123: same here

src/scripts/main.js
```diff
@@ -1,3 +1,253 @@
...
+      grid.slideLeft();
+      break;
+    case 'ArrowDown':
+      grid.slideDown();
+      break;
+    case 'ArrowRight':
+      grid.slideRight();
+      break;
+  }
+});
```
DarkMistyRoom, line 27: use object:
{
  ArrowLeft: 'ArrowLeft',
  ...
}

It's a good practice to avoid mistakes

DarkMistyRoom (changes requested):
Good job, but I've found one moment:
when there are 3 same tiles on one line and we move them to merge, the first ones should merge, not the last ones. For example, we have 3 tiles in a horizontal line and click left - the first two from the left should merge in this case

src/scripts/main.js
AnastasiiaKriuchkova: all done

sTorba24 (approved):
Well done!

PR #480, opened 2023-09-19
Develop

stunnerboyone:
[DEMO LINK](https://stunnerboyone.github.io/js_2048_game/)


stunnerboyone:
GameOver doesn't work, trying to fix

src/scripts/main.js
```diff
@@ -1,3 +1,271 @@
...
+
+    row = slide(row);
+    board[r] = row;
+  };
+
+  if (compareFields(prevBoard, board)) {
+    spawnRandomCell();
+  }
+
+  renderField();
```
etojeDenys, line 113: we can move all this logic from move functions to the end of the switch, it will simplify and remove the redundant copy logic in move functions

etojeDenys (changes requested):

sTorba24 (approved):
Looks really cool! I like it :)

PR #490, opened 2023-09-25
2048

Q1W2E3R4T5Y6U7I8a:
- [DEMO LINK](https://Q1W2E3R4T5Y6U7I8a.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,206 @@
 'use strict';
 
-// write your code here
+document.addEventListener('DOMContentLoaded', () => {
+  const gameScore = document.querySelector('.game-score');
+  const startBtn = document.querySelector('.button.start');
+  const gameCells = document.querySelectorAll('.field-cell');
+  const loseMessage = document.querySelector('.message-lose');
+  const winMessage = document.querySelector('.message-win');
+  let score = 0;
+  let board = [];
+
+  function initBoard() {
+    board = [...Array(4)].map(row => Array(4).fill(null));
+    addTile();
+    addTile();
+    renderBoard();
+    startBtn.classList.remove('start');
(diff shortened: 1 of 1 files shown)


sTorba24 (changes requested):
Generally works good

Some things to improve 
After restart score should be reset, also message 'press start` should disappear when game started, same with `you lost` when restarted

image

src/scripts/main.js
```diff
@@ -1,3 +1,213 @@
...
+    if (direction === 'down' || direction === 'right') {
+      tiles.reverse();
+    }
+
+    const mergedTiles = [];
+    let skipMerge = false;
+
+    for (const tile of tiles) {
+      if (mergedTiles.length && mergedTiles[mergedTiles.length - 1] === tile
+          && !skipMerge) {
```
loralevitska, line 128: Long expression move to variable with meaningful variable naming

src/scripts/main.js
```diff
@@ -1,3 +1,213 @@
...
+        break;
+      case 'ArrowDown':
+        move('down');
+        break;
+      case 'ArrowLeft':
+        move('left');
+        break;
+      case 'ArrowRight':
+        move('right');
+        break;
```
loralevitska, line 210:
```suggestion
      case 'ArrowUp':
        move('up');
        break;
        
      case 'ArrowDown':
        move('down');
        break;
        
      case 'ArrowLeft':
        move('left');
        break;
        
      case 'ArrowRight':
        move('right');
        break;
```

loralevitska (approved):
Good job)
remove outline, it breaks design 
image

Q1W2E3R4T5Y6U7I8a:
> Good job) remove outline, it breaks design image

image
I don't have that

PR #491, opened 2023-09-25
2048

BeginerAlex:
I made the main task, but I also wanted to add animation when the cells merged but I can't find solution to make it. I had some tries but all of them were failed. I understend that I have to check the merged cell and add some class for this cell to make animation, but in that part of code it doesn't work and it's great works in part where I compare two cells. You can see how it's work in comment part in JS file. Maybe you can show me how I can make it. Thanks for your tips.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_2048_game/)
+    - [DEMO LINK](https://begineralex.github.io/js_2048_game/)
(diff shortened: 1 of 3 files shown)


maxim2310 (changes requested):
need to test but if you want to create a full animation of moving you need give position absolute for cell and always store a position, then manipulate of it

BeginerAlex:
Unfortunately I can't find a good solution to make animation when two cells merged so I commited my clean solution.

maxim2310 (approved):
Great job overall!

PR #493, opened 2023-09-25
add task solution

ydem1:
[DEMO LINK](https://ydem1.github.io/js_2048_game/)


etojeDenys (changes requested):
https://github.com/mate-academy/js_2048_game/assets/70220667/14a7803b-89a0-4231-bc00-85236f9a724e

when we don't have any cells to merge then we do not need to create another cell

etojeDenys (changes requested):
.

PR #494, opened 2023-09-25
Add task solution

spojrzenie:


etojeDenys (changes requested):
please, provide the link in PR description

src/scripts/main.js
```diff
@@ -1,3 +1,416 @@
...
+      showBoardToUser(gameBoard);
+      isGameOver();
+      isWin();
+    } else if (left) {
+      moveLeft();
+      showBoardToUser(gameBoard);
+      isGameOver();
+      isWin();
+    }
+  }
```
etojeDenys, line 415: better to use switch case instead of multiple if else

and

let's move out from the switch case these three functions
```
 showBoardToUser(gameBoard);
      isGameOver();
      isWin();
```

etojeDenys (changes requested):

src/scripts/main.js
spojrzenie: Ok done. Thanks for advice!

etojeDenys (approved):
good job

PR #498, opened 2023-09-27
Develop

Vovchukkul:
https://Vovchukkul.github.io/js_2048_game/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_2048_game/)
+    - [DEMO LINK](https://Vovchukkul.github.io/js_2048_game/)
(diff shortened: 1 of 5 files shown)


Andriy-Shymkiv (changes requested):
you should tell user that it is game over in such case, instead I just can't move the fields, without any messages:
image

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

1. show cell even before start game and remove this block
image
2. when i click restart, game must reSTART but not go to initial value 
3. i can move in one direction all time 
image
4. need to remove
image

Andriy-Shymkiv (changes requested):
points number 2 and 3 from previous review is not fixed:
image

Andriy-Shymkiv (approved):

PR #499, opened 2023-09-27
2048

ndv20:
[DEMO LINK](https://dmytronykorovych.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -9,6 +9,13 @@ const messageWin = document.querySelector('.message-win');
 const messageLose = document.querySelector('.message-lose');
 let canMove = false;
 
+const direction = {
+  left: 'ArrowLeft',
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,248 @@
...
+    generateRandom();
+    generateRandom();
+  }
+});
+
+document.addEventListener('keydown', (e) => {
+  e.preventDefault();
+
+  switch (e.key) {
+    case 'ArrowLeft':
```
OleziO, line 39: Move all cases to the object.

Example of usage:

```
case Directions.ArrowLeft:
...
case Directions.ArrowRight:
...
etc.
```

OleziO (changes requested):

maxim2310 (approved):
Good work!!!

PR #500, opened 2023-09-28
add task solution

4xmplme:


src/scripts/main.js
```diff
@@ -1,3 +1,180 @@
...
+function move(pressedKey) {
+  const prevBoard = [...board];
+  let isReversed = false;
+
+  if (pressedKey === 'ArrowDown' || pressedKey === 'ArrowRight') {
+    isReversed = true;
+  }
+
+  switch (pressedKey) {
+    case 'ArrowUp':
```
volodymyr-soltys97, line 140: Consider will be better if you create a variable for these words and use them everywhere because now these words are duplicated

volodymyr-soltys97 (changes requested):
Great work 🔥 
Let's improve your code

PR #506, opened 2023-10-01
done

no4kar:
[DEMO LINK](https://no4kar.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,6 +6,7 @@
     <title>2048</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
+
   <body>
     <div class="container">
       <div class="game-header">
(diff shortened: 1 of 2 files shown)


anastasiiavorobiova (changes requested):
Great work! Currently, the new item is added to the game board even if there is no movement of the fields happening. Consider adding a check if it's possible to move squares before adding the new one

etojeDenys (changes requested):
https://github.com/mate-academy/js_2048_game/assets/70220667/82a75272-e763-408d-a97b-fbe88d24f550

we don't need to add a new cell if we have nothing to merge

loralevitska (approved):
Well done) Check my comments for improvement

- remove outline, it breaks design

PR #507, opened 2023-10-02
add task solution

Albert-Kifiuk:
- [DEMO LINK](https://Albert-Kifiuk.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,6 +15,7 @@
             Score: <span class="game-score">0</span>
           </p>
           <button class="button start">Start</button>
+          <button class="button restart">Restart</button>
         </div>
       </div>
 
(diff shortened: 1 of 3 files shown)


sTorba24 (changes requested):
Almost good! Lets fix a little bug you have

Game shouldn't continue when i click on button and no spaces to move cells + nothing to merge
i could press one direction button and play till the end
try to press only 'right' for example and then do the same in original game
https://2048game.com/

maxim2310 (changes requested):

1. i still can press just one direction and play 
2. when I can continue playing, you write that the game is over 
image

plodnikate (approved):
Good job on it!

PR #514, opened 2023-10-04
2048

HokageHEL:
[DEMO LINK](https://HokageHEL.github.io/js_2048_game/)

as inspiration I used this tutorial
https://www.youtube.com/watch?v=XM2n1gu4530&t=997s&ab_channel=KennyYipCoding

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,308 @@
 'use strict';
 
-// write your code here
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,308 @@
...
+    slideRight();
+  }
+
+  if (e.code === 'ArrowUp') {
+    slideUp();
+  }
+
+  if (e.code === 'ArrowDown') {
+    slideDown();
+  }
```
etojeDenys, line 130: let's use the switch case here, instead of the multiple if conditions

etojeDenys (changes requested):

denys-danyliuk (approved):

PR #517, opened 2023-10-06
Solution

olena-hapon:
[DEMO LINK](https://olena-hapon.github.io/js_2048_game/)


maxim2310 (changes requested):
Good work!!!
But need to check one point, now i can click in one direction all time even when i shouldn't

maxim2310 (approved):
Good work!!!

PR #518, opened 2023-10-07
add task solution

natamrshn:


etojeDenys (changes requested):
We shouldn't add the new cell if we don't have any cells to merge
https://github.com/mate-academy/js_2048_game/assets/70220667/a17c3e58-e2b3-4b7a-8781-fbd6175c2bf5

volodymyr-soltys97 (approved):
Well done 🔥

PR #526, opened 2023-10-15
implementation

Daniilart01:


volodymyr-soltys97 (changes requested):
Nice work 👍 
By default there should be no numbers, they should appear after pressing start, and also the text below should disappear after pressing start
image

PR #527, opened 2023-10-15
add boilerplate and need to check exports and privat #

SerhiiUnhurian:
- [DEMO LINK](https://SerhiiUnhurian.github.io/js_2048_game/)

### DIFF
--- a/.babelrc 
+++ b/.babelrc 
@@ -1,0 +1,11 @@
+{
+  "presets": [
+      "@babel/preset-env",
+      "@babel/preset-react"
+  ],
+  "plugins": [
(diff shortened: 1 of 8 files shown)


maxim2310 (changes requested):
but next time when something went wrong ask in chat
image
image

IvanFesenko (changes requested):
Demo Links isn't working
If you have any questions - ask them in fe_chat

SerhiiUnhurian:
that's why i sent it here, there is a problem with private variables, and i don't know how to fix it

PR #532, opened 2023-10-18
add task solution

alina-kabanets:
https://alina-kabanets.github.io/js_2048_game/


sTorba24 (changes requested):
Good work! lets fix mismatch with original game
when nothing to merge and no space to move cells game shouldn't continue
try to click only 'right' button for example and do the same in original game https://2048game.com/

VitaliyBondarenko1982 (approved):
Well done.

PR #542, opened 2023-10-23
solution

NataliMax:


NataliMax:
[DEMO LINK](https://NataliMax.github.io/js_2048_game/)

anastasiiavorobiova (changes requested):
Great start! Keep doing: arrows don't work, a user can just start or restart the game. Feel free to ask for help in the chat

NataliMax:
It`s work with 'a' ,'d', 'w', 's'. please try

denys-danyliuk (changes requested):
problem about which I wrote earlier still exists
Знімок екрана 2023-10-24 о 09 39 31

src/scripts/main.js
```diff
@@ -1,3 +1,406 @@
...
+      break;
+    }
+
+    case 's': {
+      moveVertically(down);
+      break;
+    }
+  }
+  loseCheck();
+});
```
sTorba24, line 256: add also ability to use arrows, will be really good improvement

src/scripts/main.js
```diff
@@ -1,3 +1,406 @@
...
+//         break;
+//       default: return 0;
+//     }
+//     loseCheck();
+//   }
+// });
+
+// document.addEventListener('touchmove', (e) => {
+//   e.preventDefault();
+// });
```
sTorba24, line 327: remove commented code

sTorba24 (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,363 @@
...
+  switch (e.key) {
+    case 'a': {
+      moveHorizontally(left);
+      break;
+    }
+
+    case 'ArrowLeft': {
+      moveHorizontally(left);
+      break;
+    }
```
etojeDenys, line 243: we can combine these cases
```suggestion
    case 'a': 
    case 'ArrowLeft': {
      moveHorizontally(left);
      break;
    }
```

etojeDenys (changes requested):

src/scripts/main.js
NataliMax: Unfortunately, the linter requires a 'break' after each 'case'. I asked this question in the chat, In this case it is impossible

etojeDenys (approved):
lgtm

PR #550, opened 2023-10-27
I am having issues with the num generator not always seeing that a sq…

spencerspe:
…uare is filled. The rest of the code is still being worked on
    - [DEMO LINK](https://spencerspe.github.io/js_2048_game/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_2048_game/)
+    - [DEMO LINK](https://spencerspe.github.io/js_2048_game/)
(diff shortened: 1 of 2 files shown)


anastasiia-tilikina (changes requested):
We've discussed the problem in chat

anastasiia-tilikina (changes requested):
Good start!

but we have some bugs here)

1. field shouldn't have any cells on page loading
2. cells should appear after start button click
3. when game is not started pressing keys should cause nothing
4. I can press one button (for example 'down') new cells will always appear. But they should appear only if something changed on the move (cells were moved or merged)
5. the game is over but there are still options for moves
image

lerastarynets (approved):
lgtm 🚀

PR #552, opened 2023-10-29
solution

YuriiHlushenko:
- [DEMO LINK](https://YuriiHlushenko.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,390 @@
 'use strict';
 
-// write your code here
+const start = document.querySelector('.start');
+const score = document.querySelector('.game-score');
+const messageStart = document.querySelector('.message-start');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+
+score.textContent = 0;
+
+const cells = {
+  1: '',
+  2: '',
+  3: '',
+  4: '',
+  5: '',
+  6: '',
+  7: '',
+  8: '',
+  9: '',
+  10: '',
+  11: '',
+  12: '',
+  13: '',
+  14: '',
+  15: '',
+  16: '',
+};
+
+const colorMods = [
+  'field-cell--2',
+  'field-cell--4',
+  'field-cell--8',
+  'field-cell--16',
+  'field-cell--32',
+  'field-cell--64',
+  'field-cell--128',
+  'field-cell--256',
+  'field-cell--512',
+  'field-cell--1024',
+  'field-cell--2048',
+];
+
+function getRandom(min, max) {
+  return Math.round((min - 0.5) + Math.random() * (max - min + 1));
+}
+
+function removeColors() {
+  for (const key in cells) {
+    for (const mod of colorMods) {
+      if (document.querySelector(`.cell_${key}`).classList.contains(mod)) {
+        document.querySelector(`.cell_${key}`).classList.remove(mod);
+      }
+    }
+  }
+}
+
+function sync() {
+  for (const key in cells) {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,390 @@
...
+      }
+    }
+  }
+}
+
+function sync() {
+  for (const key in cells) {
+    document.querySelector(`.cell_${key}`).textContent = cells[key];
+
+    switch (cells[key]) {
```
TarasHoliuk, line 62: what about smth like

```
if (cells[key]) {
  document.querySelector(`.cell_${key}`).classList.add(`field-cell--${cells[key]}`)
}
```
instead of the entire switch block (you can save cells[key] in variable)

src/scripts/main.js
```diff
@@ -1,3 +1,390 @@
...
+  'field-cell--512',
+  'field-cell--1024',
+  'field-cell--2048',
+];
+
+function getRandom(min, max) {
+  return Math.round((min - 0.5) + Math.random() * (max - min + 1));
+}
+
+function removeColors() {
```
TarasHoliuk, line 48: you can query all cells, iterate through the list, and set the default class for each
and get rid of `colorMods ` array

src/scripts/main.js
```diff
@@ -1,3 +1,390 @@
...
+        break;
+    }
+  }
+}
+
+function addNumbers() {
+  let cellNum1 = getRandom(1, 16);
+  let cellNum2 = getRandom(1, 16);
+
+  while (cells[cellNum1] !== '') {
```
TarasHoliuk, line 106:
```suggestion
  while (cells[cellNum1]) {
```
empty string is a [falsy ](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)value

src/scripts/main.js
```diff
@@ -1,3 +1,390 @@
...
+    cellNum1 = getRandom(1, 16);
+  }
+
+  while (cells[cellNum2] !== '' || cellNum1 === cellNum2) {
```
TarasHoliuk, line 110: to avoid such things here: `cellNum1 === cellNum2` refactor the function: make it add only one number in an empty cell. Then. if you need to add 2 numbers call this function 2 times

TarasHoliuk: Oh, I see you have such below - just get rid of `addNumbers`

src/scripts/main.js
```diff
@@ -1,3 +1,390 @@
...
+    if (checkCol() === false) {
+      messageLose.classList.remove('hidden');
+    }
+  }
+}
+
+document.addEventListener('keydown', e => {
+  switch (e.key) {
+    case 'ArrowUp':
+      if (!messageStart.classList.contains('hidden')) {
```
TarasHoliuk, line 337: FYI: you may add this listener on `Start` button click and remove this condition

src/scripts/main.js
```diff
@@ -1,3 +1,390 @@
...
+        break;
+      }
+      moveUp();
+      sumUp();
+      addNumber();
+      removeColors();
```
TarasHoliuk, line 343: instead of removing all classes and adding those that are required in the next function, try to modificate classes in the proper way inside `sycnc` function (optinally)

TarasHoliuk (changes requested):
GJ 👍  Added a few comments related to code
Also, 1 requirement is necessary: do not add new cell if after arrow button click nothing changed (no one cell was moved or merged)

sTorba24 (approved):

PR #553, opened 2023-10-30
Task solution

kseniia-chepur:
[DEMO LINK](https://kseniia-chepur.github.io/js_2048_game/)


maxim2310 (changes requested):
1. i can click only in one direction even if if i shouldnt 
2. i have a loose game when i had a unmerged cell
image
3. after restart i cant move

volodymyr-soltys97 (approved):
Well done 🔥

PR #557, opened 2023-11-03
solution

maksym-kostetskyi:
https://maksym-kostetskyi.github.io/js_2048_game/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_2048_game/)
+    - [DEMO LINK](https://maksym-kostetskyi.github.io/js_2048_game/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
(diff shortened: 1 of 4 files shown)


VitaliyBondarenko1982 (changes requested):
Please provide demo link in PR description.
Also suggest to make some refactoring.
- write magic numbers in variables. looks like it can be one variable something like const GAME_SIZE = 4 and use it in all places to replace 4 , 16, 3 hardocoded numbers.
- move function looks like rather similar. try to group same or similar logic in functions. it can reduce amount of code and make code more cleaner

VitaliyBondarenko1982 (changes requested):
Thanks for demo link.
Found a couple bugs.
- after start game should appear two cells wiht number.
- after any move if no cells for merge and any cell wasn't moved - new cell shouldn't appear.

maxim2310 (changes requested):
find one bug, please fix it before approval 
restart must start new game immediately

maxim2310 (approved):
Good job!!!

PR #559, opened 2023-11-04
js_2048_game

Donovanoff:
[DEMO LINK](https://Donovanoff.github.io/js_2048_game/)


sTorba24 (changes requested):
Almost good! Lets improve it

game shouldn't continue when click on button won't change something(merge cells or move them)

try to click one-direction button('right` for example) and check how it is
on board nothing changing but new items still appears

check then original game how it should be https://2048game.com/

src/scripts/main.js
```diff
@@ -1,3 +1,391 @@
...
+
+  if (e.code === 'ArrowLeft') {
+    slideLeft();
+  } else if (e.code === 'ArrowRight') {
+    slideRight();
+  } else if (e.code === 'ArrowUp') {
+    slideUp();
+  } else if (e.code === 'ArrowDown') {
+    slideDown();
+  }
```
SanyaBratashchuk, line 66: use switch case

SanyaBratashchuk: also move strings to constants

SanyaBratashchuk (changes requested):

TarasHoliuk (approved):

PR #570, opened 2023-11-11
added game 2048

MartaChobaniuk:
[DEMO LINK](https://martachobaniuk.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,175 @@
 'use strict';
 
-// write your code here
+const startMessage = document.querySelector('.message-start');
+const winMessage = document.querySelector('.message-win');
+const loseMessage = document.querySelector('.message-lose');
+
+const gameScore = document.querySelector('.game-score');
+const startButton = document.querySelector('.button');
+const fieldRows = document.querySelectorAll('.field-row');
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,175 @@
...
+function fillFieldCells() {
+  fillRandomCell();
+  fillGameField();
+}
+
+function handleArrows(e) {
+  switch (e.key) {
+    case 'ArrowLeft':
+      clickArrowLeft();
+      fillFieldCells();
```
etojeDenys, line 139: you can move this out of the switch and call only one time (after the switch)

etojeDenys (changes requested):
1. styles here look broken
image

2. you don't need to add another cell, if we don't merge any other cells

https://github.com/mate-academy/js_2048_game/assets/70220667/d5244ec4-b9a8-4d5e-9986-43a26c65d5fa

etojeDenys (changes requested):
image
still can reproduce this bug

MartaChobaniuk:
I don't see this bug

lerastarynets (approved):

PR #574, opened 2023-11-12
made solution

MrHunterBoi:
Somehow it can't find the startGame() function. Everything works fine on my side though


polosanya (changes requested):
Please don't forget to deploy your work and add demo link to the PR description so we can properly check it 🙏

MrHunterBoi:
Oh, right. Forgot to update the comment after fixing stuff xd

PR #577, opened 2023-11-15
game2048

MariiaSuper:
https://MariiaSuper.github.io/js_2048_game/


maxim2310 (changes requested):
now i can click in in one direction(right for example) even if i shouldnt and blocks appears, check it and fix please

xiimiik (approved):
nice work!

PR #586, opened 2023-11-21
Develop

RomanHrytsiuk:
https://RomanHrytsiuk.github.io/js_2048_game/


src/scripts/main.js
```diff
@@ -1,3 +1,344 @@
...
+  newRow = clearZeroes(newRow);
+
+  while (newRow.length < columns) {
+    newRow.push(0);
+  }
+
+  return newRow;
+}
+
+function slideLeft() {
```
anastasiia-tilikina, line 198: seems like slideLeft and slideRight differ only by reversed array, let's try to join these two functions into one for not writing the same code twice

src/scripts/main.js
```diff
@@ -1,3 +1,344 @@
...
+      updateFieldCell(fieldCell, cellValue);
+    }
+  }
+
+  if (isAnyCellMoved) {
+    addRandomCell();
+  }
+}
+
+function slideUp() {
```
anastasiia-tilikina, line 255: same with slideUp slideDown

src/scripts/main.js
```diff
@@ -1,3 +1,344 @@
...
+        move = false;
+      } else if (c < columns - 1 && board[r][c] === board[r][c + 1]) {
+        move = false;
+      } else if (r < rows - 1 && board[r][c] === board[r + 1][c]) {
+        move = false;
+      } else if (r > 0 && board[r][c] === board[r - 1][c]) {
+        move = false;
+      }
+    }
+  }
```
anastasiia-tilikina, line 329: instead of writing else if use keyword break to exit loop in each if brunch. It will keep you code clean.

```javascript
if (condition) {
move = false;
break;
}

if (condition) {
move = false;
break;
}

// and so on
``` 

or you can even combine those conditions since you're doing the same action.

```javascript
//moving conditions into variables for redability
const condition1 = // your code;
const condition2 = // your code;

if (condition1 || condition2 || //...) {
move = false;
}
```

anastasiia-tilikina (changes requested):
Very well done! 🔥 

let's just improve your code style)

anastasiiavorobiova (approved):
Great work!

PR #587, opened 2023-11-22
Develop

boytrend0108:
[DEMO LINK](https://boytrend0108.github.io/js_2048_game/)


Esceype (changes requested):
GJ!
But it`s works a bit wrong.
If a side already has 4 items in a column or row and there is nothing to merge, the player should not be able to press the button that is responsible for that side again. For example, we have 4 blocks in a column on the right side and there is a new block 2 with nothing to merge with, so pressing the key to the right side should be ignored.

xiimiik (changes requested):
Nice work!
Overall fix the score, it should be like on the reference - the score is added when the numbers add up, not from the maximum block, as it is now

volodymyr-soltys97 (approved):
Great work 🔥

PR #589, opened 2023-11-23
Solution

garmoni:


lerastarynets (changes requested):
Pls deploy your project and add a link to it in the pr

xiimiik (changes requested):
GJ!
But it's works a little bit wrong.

1. If a side already has 4 items in a column or row and there is nothing to merge, the player should not be able to press the button that is responsible for that side again. For example, we have 4 blocks in a column on the right side and there is a new block 2 with nothing to merge with, so pressing the key to the right side should be ignored.
2. On main screen, when I press "Start" button, and and the text "Press start to begin..."  disappears, the field is "jumping"
3. Unnecessary overlay under the bottom of field
Снимок экрана 2023-11-23 в 12 29 33

IvanFesenko (approved):
well done

PR #599, opened 2023-11-30
Dont work npm run deploy... :(

GlibVK:


.github/workflows/test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
```
TarasHoliuk, line 14: does somebody from mentors help you to change settings?

TarasHoliuk (changes requested):
I cannot be sure why the `deploy` doesn't work. Probably it cannot make the correct build (strange that `npm start` works). Ask help in the chat or go to Q&A session

PR #600, opened 2023-11-30
add task solution

heorhiig:


sTorba24 (changes requested):
good work!

lets improve it

1. game should be initialized with 2 cells from start
image

2. cell shouldn't be added if on current board nothing to merge and no space to move
try to click only 'right' button and do the same in original game https://2048game.com/

note the difference

3.remove message after start
image


4.there are cells to merge, so game isn't end
image

5. Add restart button
image

src/scripts/main.js
```diff
@@ -1,3 +1,282 @@
...
+  }
+
+  let found = false;
+
+  while (!found) {
+    const r = Math.floor(Math.random() * rows);
+    const c = Math.floor(Math.random() * columns);
+
+    if (board[r][c] === 0) {
+      const num = Math.random() < 0.01 ? 4 : 2; // Fix the missing parentheses
```
loralevitska, line 98: Remove comment. Try to write obvious code with meaningful variable and function naming

loralevitska (approved):
Almost perfect) 
Add a random number generation so that either 2 or 4 is added, because otherwise this is the picture
image

maxim2310 (changes requested):
Still same problem and would be better if game will start after click on start button 
image

PR #613, opened 2023-12-08
Solution

oksana-vas:
[DEMO LINK](https://oksana-vas.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,244 @@
 'use strict';
 
-// write your code here
+const startButton = document.querySelector('.button');
+const gameField = document.querySelector('.game-field');
+const gameScore = document.querySelector('.game-score');
+const cells = document.querySelectorAll('.field-cell');
+
+const messages = document.querySelectorAll('.message');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+
+let score = 0;
+
+const BOARD_SIZE = 4;
+const SCORE_TO_WIN = 2048;
+
+const board = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
+
+startButton.addEventListener('click', () => {
+  if (startButton.classList.contains('start')) {
+    startGame();
+  }
+
+  [...messages].map(message => message.classList.add('hidden'));
+
+  for (let i = 0; i < BOARD_SIZE; i++) {
+    for (let j = 0; j < BOARD_SIZE; j++) {
+      board[i][j] = 0;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,264 @@
...
+  }
+}
+
+function updateScore(value) {
+  score += value;
+  gameScore.textContent = score;
+}
+
+function isWinner() {
+  return [...cells].some(cell => {
```
VitaliyBodnar, line 124: redundant 
```suggestion
  return cells.some(cell => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,264 @@
...
+  addRandomNumber();
+  addRandomNumber();
+
+  renderBoard();
+});
+
+// eslint-disable-next-line no-shadow
+document.addEventListener('keydown', (event) => {
+  switch (event.key) {
+    case 'ArrowUp':
```
VitaliyBodnar, line 54: ArrowUp, ArrowDown, ArrowRight and ArrowLeft are magic words. Consider will be better if you create a constants  for magic words and use it everywhere

src/scripts/main.js
```diff
@@ -1,3 +1,264 @@
...
+  for (let i = 0; i < BOARD_SIZE; i++) {
+    for (let j = 0; j < BOARD_SIZE; j++) {
+      board[i][j] = 0;
+    }
+  }
+
+  [...cells].map(cell => {
+    cell.className = 'field-cell';
+    cell.textContent = '';
+  });
```
VitaliyBodnar, line 40: map is used when you need to return a new array. In this case it's make sense to use forEach here.
```suggestion
  cells.forEach(cell => {
    cell.className = 'field-cell';
    cell.textContent = '';
  });
```

src/scripts/main.js
oksana-vas: but we need to gather cells into array. without square brackets it doesn't work
Знімок екрана 2023-12-08 о 13 47 22

src/scripts/main.js
```diff
@@ -1,3 +1,269 @@
 'use strict';
 
-// write your code here
+const startButton = document.querySelector('.button');
+const gameField = document.querySelector('.game-field');
+const gameScore = document.querySelector('.game-score');
+const cells = document.querySelectorAll('.field-cell');
```
IvanFesenko, line 6: Just comment: instead of use spread in few places you can make cell prepared before
```suggestion
const cells = [...document.querySelectorAll('.field-cell')];
```

IvanFesenko (approved):
Well done

src/scripts/main.js
oksana-vas: thank you! so simple but really useful :)

PR #632, opened 2023-12-18
new commit

Max-just:
- [DEMO LINK](https://Max-just.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,248 @@
 'use strict';
 
-// write your code here
+window.onload = function() {
+  let score = 0;
+  const rows = 4;
+  const columns = 4;
+  let board = [];
+  let gameStarted = false;
+
+  const gameField = document.querySelector('.game-field');
+  const gameScore = document.querySelector('.game-score');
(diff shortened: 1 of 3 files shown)


maxim2310 (changes requested):
Yor demo link is not working, please fix it and make request review again

loralevitska (changes requested):
Good job! There are things that should be fixed:

- these pictures funny for Ukranians but don't understandable for others. This work will be on resume, so please change background without images
image
- add random value for empty cell (2 or 4), now you have only 2 number

src/scripts/main.js
```diff
@@ -1,3 +1,253 @@
 'use strict';
 
-// write your code here
+window.onload = function() {
+  let score = 0;
+  const rows = 4;
+  const columns = 4;
+  let board = [];
+  let gameStarted = false;
```
Andriy-Shymkiv, line 8:
```suggestion
  let isGameStarted = false;
```

Andriy-Shymkiv (approved):
very good!

PR #638, opened 2023-12-24
Develop

petrolozynskyi:
https://petrolozynskyi.github.io/js_2048_game/


src/scripts/main.js
```diff
@@ -1,3 +1,273 @@
...
+  }
+
+  function clearStyles() {
+    for (let i = 0; i < squares.length; i++) {
+      squares[i].style.backgroundColor = '';
+    }
+  }
+
+  function handleNewGameClick() {
+    location.reload();
```
anastasiia-tilikina, line 30: why do you reload page on resetting game? You can avoid it

src/scripts/main.js
```diff
@@ -1,3 +1,273 @@
...
+    while (squares[randomNumber].innerHTML !== '0') {
+      randomNumber = Math.floor(Math.random() * squares.length);
+    }
+
+    squares[randomNumber].innerHTML = '2';
+    checkForGameOver();
+  }
+
+  function moveRight() {
+    for (let i = 0; i < 16; i++) {
```
anastasiia-tilikina, line 64: 16 should be saved in some variable,  so we can avoid using [magic numbers](https://www.pluralsight.com/tech-blog/avoiding-magic-numbers/) look through your code and fix other cases

src/scripts/main.js
```diff
@@ -1,3 +1,273 @@
...
+    if (e.keyCode === 37) {
+      keyLeft();
+    } else if (e.keyCode === 38) {
+      keyUp();
+    } else if (e.keyCode === 39) {
+      keyRight();
+    } else if (e.keyCode === 40) {
+      keyDown();
+    }
+  }
```
anastasiia-tilikina, line 185: it's better to use switch case structure here

anastasiia-tilikina (changes requested):
1. your animation seems to be glitched now. for one moment cell show '0' number after I move.
2. If I press some button (e.g. 'down') multiple times new cells keep appear even if nothing was changed. New cell should appear only if there were some changes (cells were moved or merged)
3. I still have moves but the game is lost.
image

4. your functions move(right/up/down/left) share some logic. try to refactor your code and avoid duplicates. As well as combineRow and combineColl fun.

src/scripts/main.js
etojeDenys: image

etojeDenys (changes requested):

maxim2310 (changes requested):
one things left, in 10% cases must appear block with 4

maxim2310 (approved):
Well done

PR #640, opened 2023-12-25
Develop

Patryk-Buczkowski:
- [DEMO LINK](https://Patryk-Buczkowski.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -113,7 +113,7 @@ window.addEventListener('load', () => {
 
     move({ moveDirectionCallback, xIsWithinBoard, yIsWithinBoard,
       nextIndexY, columnStartIndex, nextIndexX,
-      rowStartIndex, currentX, currentY, mergedCells = [] }) {
+      rowStartIndex, currentX, currentY }) {
       if (xIsWithinBoard) {
         if (yIsWithinBoard) {
           const temp = this.board[currentY][currentX];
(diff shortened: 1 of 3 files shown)


volodymyr-soltys97 (changes requested):
Hi, your demo link doesn't work, you need to make a deploy
image

Patryk-Buczkowski:
When I type npm run deploy vs code show error

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
At this stage, it seems that the page is frozen and nothing happens, if the user has lost, then a message about this should be shown, but there should still be an opportunity to continue playing
image

volodymyr-soltys97 (changes requested):
To improve:
Now the message at the bottom about the loss disappears after clicking the arrow, then reappears and disappears, this needs to be fixed so that it does not disappear
image
image

volodymyr-soltys97 (changes requested):
Need to remove the losing text after we start a new game
image

src/scripts/main.js
```diff
@@ -1,3 +1,254 @@
...
+
+    if (ev.key === 'ArrowUp') {
+      ev.preventDefault();
+      htmlBoard.moveUp();
+    }
+
+    if (ev.key === 'ArrowDown') {
+      ev.preventDefault();
+      htmlBoard.moveDown();
+    }
```
volodymyr-soltys97, line 222: Use a switch case here instead of if

src/scripts/main.js
```diff
@@ -1,3 +1,254 @@
...
+  }
+
+  const htmlBoard = new Board();
+
+  document.addEventListener('keydown', ev => {
+    if (ev.key === 'ArrowLeft') {
+      htmlBoard.moveLeft();
+    }
+
+    if (ev.key === 'ArrowRight') {
```
volodymyr-soltys97, line 210: Will be better if you create a variables for these words and use it here

src/scripts/main.js
```diff
@@ -1,3 +1,254 @@
...
+      ];
+
+      messageLose.classList.add('hidden');
+      htmlBoard.appendOneBox();
+      htmlBoard.appendOneBox();
+      htmlBoard.score = 0;
+      scoreElem.textContent = '0';
+    }
+
+    buttonStart.textContent = 'Restart';
```
volodymyr-soltys97, line 252: The same as above

volodymyr-soltys97 (changes requested):
Great job 👍 
Let's improve your code

src/scripts/main.js
Patryk-Buczkowski: i don't understand this suggestion and thin one below

anastasiiavorobiova (changes requested):
Great work! 
A new number is added even if the game field is not changed. 
Also, the new number should move with the old values on the key press, consider fixing it

anastasiiavorobiova (changes requested):
Great progress here! Unfortunately, new digits are added even if the game field doesn't change. A user can't move even if there are available moves here 
Screenshot 2023-12-30 at 18 32 04

src/scripts/main.js
```diff
@@ -1,3 +1,310 @@
...
+        ev.preventDefault();
+        htmlBoard.moveUp();
+        break;
+      case arrowDown:
+        ev.preventDefault();
+        htmlBoard.moveDown();
+        break;
+      default:
+        throw new Error('wrong button');
+    }
```
volodymyr-soltys97, line 277:
```suggestion
    switch (ev.key) {
      case arrowLeft:
        htmlBoard.moveLeft();
        break;
        
      case arrowRight:
        htmlBoard.moveRight();
        break;
        
      case arrowUp:
        ev.preventDefault();
        htmlBoard.moveUp();
        break;
        
      case arrowDown:
        ev.preventDefault();
        htmlBoard.moveDown();
        break;
        
      default:
        throw new Error('wrong button');
    }
```

volodymyr-soltys97 (approved):
Great work 🔥

PR #646, opened 2024-01-01
done

Aleksey-Aleksey:
[DEMO LINK](https://Aleksey-Aleksey.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,232 @@
 'use strict';
 
-// write your code here
+const page = document.documentElement;
+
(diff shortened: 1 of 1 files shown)


maxim2310 (changes requested):
now block appear even if no one movement done, in this case if i click down blocks will not move but new block will be appear
image

Aleksey-Aleksey:
> Good work, but lets make some improvement now block appear even if no one movement done, in this case if i click down blocks will not move but new block will be appear image

please check again

maxim2310 (approved):
Well done!!!

PR #650, opened 2024-01-03
Add task solution

dimakrugly:
[DEMO LINK](https://dimakrugly.github.io/js_2048_game/)


maxim2310 (approved):
Good work, just check i think that block with 4 appear very often(it must be about 10%)

PR #655, opened 2024-01-09
Add solution task

Eater228:
- [DEMO LINK](https://Eater228.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,181 @@
 'use strict';
 
-// write your code here
+const startMessage = document.querySelector('.message-start');
+const winnerMessage = document.querySelector('.message-win');
+const loseMessage = document.querySelector('.message-lose');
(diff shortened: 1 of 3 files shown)


lerastarynets (changes requested):
It seems that I can't lose, nothing happens when the user can't make any more moves
Screenshot at Jan 09 12-22-57

src/scripts/main.js
```diff
@@ -1,3 +1,188 @@
...
+function fillAllCels() {
+  createRandomNumb();
+  changedGameField();
+}
+
+function pushedArrows(element) {
+  const originalField = JSON.stringify(field);
+
+  switch (element.key) {
+    case 'ArrowLeft':
```
volodymyr-soltys97, line 140: You can create a variable for this word and use it here

volodymyr-soltys97 (approved):
Nice work 🔥

PR #656, opened 2024-01-10
add task solution

isaeev:


maxim2310 (changes requested):

1. when i dont press arrow and blocs didn move, new block must not appear(you could just press down and will see)
2. actually am not sure but when i press arrow down 10 times a row, then applet not react on my input and looks like you have infinity loop somewhere because in console i havent any error, maybe its my browser, so check it plz

etojeDenys (changes requested):
looks like your demo link is broken, feel free to ask about that in the chat

loralevitska (approved):
Well done)

PR #659, opened 2024-01-13
add task solutions

MrWhiteres:
[DEMO LINK](https://MrWhiteres.github.io/js_2048_game/)


SanyaBratashchuk (changes requested):
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
When you press the left or right arrow several times, a new square does not appear, although there is a place and it should appear. This needs to be fixed
image
image

lerastarynets (approved):

PR #670, opened 2024-01-18
add solution 2048, ready to play

Diliradon:
https://Diliradon.github.io/js_2048_game/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
We need to remove the text at the bottom about losing, after we start a new game
image

anastasiiavorobiova (approved):
Great work!

PR #671, opened 2024-01-19
Develop

yurovych:
https://yurovych.github.io/js_2048_game/
There is not implemented smooth movements of tiles
Mentor said it's now nesseseraly


l4st1m0za (approved):
Great job!

PR #674, opened 2024-01-22
[Logic tests]: add tests to check game logic

sergii-nosachenko:
## Description

This pull request adds the basic structure of the Game class and logic tests to verify it. Additionally, it introduces a custom Jest Mochawesome reporter that produces a proper JSON report.

### Screenshots

Mochawesome-Report


README.md
```diff
@@ -8,7 +8,8 @@ Don't play for too long! We need you to write the code!
 
 Okay, what do we have?
 1) HTML and CSS are already written. You can use it, or implement your own design if you want.
-2) Reference.
+2) Base `Game` class structure is already written too. Extend it with your own methods.
```
mgrinko, line 11: please also expain that tests check the code using ... methods
Otherwise it may not be clear that they should keep them

src/modules/Game.class.js
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * This class represents the game.
+ * Now it has a basic structure, that is needed for testing.
+ * Feel free to extend it as you wish.
+ */
+class Game {
+  /**
+   * Creates a new game instance.
+   * @param {Array<Array<number>>} initialState
```
mgrinko, line 11:
```suggestion
   * Creates a new game instance.
   *
   * @param {Array<Array<number>>} initialState
```

src/modules/Game.class.js
```diff
@@ -0,0 +1,75 @@
...
+   * Movement methods
+   */
+  moveLeft() {}
+  moveRight() {}
+  moveUp() {}
+  moveDown() {}
+
+  /**
+   * Returns the current game score.
+   * @returns {number} The current game score.
```
mgrinko, line 46: You don't need to mention it twice

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+        expect(typeof game2048[method]).toBe('function');
+      });
+    });
+  });
+
+  describe('initial state', () => {
+    it('should have a state of an empty board by default', () => {
+      const game2048 = new Game();
+
+      const state = game2048.getState();
```
mgrinko, line 43:
```suggestion
      const game2048 = new Game();
      const state = game2048.getState();
```

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+
+      expect(game2048.getStatus()).toBe('idle');
+    });
+
+    it('should have a state of a custom board if the initial state is provided', () => {
+      const initialState = [
+        [2, 0, 0, 0],
+        [0, 4, 0, 0],
+        [0, 0, 8, 0],
+        [0, 0, 0, 16]];
```
mgrinko, line 70:
```suggestion
        [0, 0, 0, 16],
      ];
```

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+
+      expect(state).toEqual(initialState);
+    });
+
+    it('should have a score of a custom value if the initial score is provided', () => {
+      const initialScore = 42;
+
+      const game2048 = new Game(null, initialScore);
+
+      expect(game2048.getScore()).toBe(initialScore);
```
mgrinko, line 84:
```suggestion
      const game2048 = new Game(null, 42);

      expect(game2048.getScore()).toBe(42);
```

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+        [0, 0, 0, 16]];
+
+      const game2048 = new Game(initialState);
+
+      const state = game2048.getState();
+
+      expect(state).toEqual(initialState);
+    });
+
+    it('should have a score of a custom value if the initial score is provided', () => {
```
mgrinko, line 79: Do you think we really should test it?

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+      const initialScore = 42;
+
+      const game2048 = new Game(null, initialScore);
+
+      expect(game2048.getScore()).toBe(initialScore);
+    });
+
+    it('should have a status of a custom value if the initial status is provided', () => {
```
mgrinko, line 87: I don't think a custom status is really important

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+
+    it('should have a status of "playing" after start', () => {
+      const game2048 = new Game();
+
+      game2048.start();
+
+      expect(game2048.getStatus()).toBe('playing');
+    });
+
+    it('should have a state with two random cells with 2 or 4 after start', () => {
```
mgrinko, line 113: the original game generate 4 in 20% of cases otherwise 2
So i would just test that there are 2 values <= 4

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+      game2048.start();
+
+      const flatState = game2048.getState().flat();
+
+      expect(
+        flatState.filter((cell) => [2, 4].includes(cell)),
+      ).toHaveLength(2);
+    });
+
+    it('should initialize the board with 4s less often than with 2s', () => {
```
mgrinko, line 125: no need in such a detailed test
It is not so important

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+      [8, 8, 2, 2],
+      [8, 8, 2, 2],
+    ];
+
+    const initialStateMergingAndStacking = [
+      [2, 2, 0, 2],
+      [0, 0, 2, 2],
+      [4, 2, 2, 0],
+      [4, 0, 0, 2],
+    ];
```
mgrinko, line 190: it is better to have these boards inside the tests using them

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+
+    describe('moveLeft', () => {
+      it('should not move the board if the game is not in the "playing" state', () => {
+        const game2048 = new Game(initialStateNoStacking, 0, 'win');
```
mgrinko, line 194: it is hard to understand the test becase its start position is not in the test

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+        const game2048 = new Game(initialStateNoStacking, 0, 'playing');
+
+        game2048.moveLeft();
+
+        const state = game2048.getState();
+
+        expect(state[0][0]).toBe(2);
+        expect(state[1][0]).toBe(4);
+        expect(state[2][0]).toBe(8);
+        expect(state[3][0]).toBe(16);
```
mgrinko, line 213: I think `getColumn(index)` helper can be really helpful here

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+        game2048.moveRight();
+
+        const flatState = game2048.getState().flat();
+
+        expect(
+          flatState
+            .some((cell, index) => (
+              index % 4 !== 3 && [2, 4].includes(cell)
+            )),
+        ).toBe(true);
```
mgrinko, line 343: I think that we can test the numbers of not empty cells 
and the total sum of numbers 

then we can just check that 1 non-empty cell was added and total sum increase <= 4

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+        expect(state[3][0]).toBe(2);
+        expect(state[3][1]).toBe(4);
+        expect(state[3][2]).toBe(8);
+        expect(state[3][3]).toBe(16);
+
+        expect(game2048.getScore()).toBe(0);
+      });
+
+      it('should add a random cell with 2 or 4 at an empty position after the move', () => {
+        const game2048 = new Game(initialStateNoStacking, 0, 'playing');
```
mgrinko, line 557: I think we can just run the `.start` method instead of passing score and status

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+
+      game2048.moveLeft();
+      game2048.moveRight();
+      game2048.moveUp();
+      game2048.moveDown();
+
+      game2048.restart();
+
+      expect(game2048.getScore()).toBe(0);
+      expect(game2048.getStatus()).toBe('idle');
```
mgrinko, line 714: I don't see that we described the expected reset logic in the task description
(I think it is better to keep it the same as in the original game)

tests/Game.class.test.js
```diff
@@ -0,0 +1,724 @@
...
+
+      game2048.moveRight();
+
+      const state = game2048.getState();
+
+      expect(game2048.getStatus()).toBe('win');
+      expect(state[1][3]).toBe(2048);
+    });
+
+    it('should be a "playing" after player has moved and board is full but new move is still possible with added tile', () => {
```
mgrinko, line 660:
```suggestion
    it('should be a "playing" after player has moved and board is full but new move is still possible with an added tile', () => {
```

mgrinko (changes requested):

jest-mochawesome.js
```diff
@@ -0,0 +1,186 @@
...
+      `${outputDir}/${outputName}.json`,
+    );
+  }
+
+  _buildElapsedTime(suites) {
+    return suites.reduce((sum, _suite) => {
+      return sum + _suite.testResults.reduce((_sum, _test) => {
+        return _sum + _test.duration;
+      }, 0);
+    }, 0);
```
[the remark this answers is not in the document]
sergii-nosachenko: Outdated, this is now a separate npm package

tests/Game.class.test.js
sergii-nosachenko: Of course, it is critical for other tests

tests/Game.class.test.js
sergii-nosachenko: Why not? We need to check if there are always more 2s than 4s, as the task requires.

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+
+    it('should have a state of a custom board if the initial state is provided', () => {
+      const initialState = [
+        [2, 0, 0, 0],
+        [0, 4, 0, 0],
+        [0, 0, 8, 0],
+        [0, 0, 0, 16],
+      ];
+
+      const game2048 = new Game(cloneState(initialState));
```
mgrinko, line 78:
```suggestion
      const game2048 = new Game(initialState.map(row => [...row]);
```

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+      game2048.start();
+
+      const flatState = game2048.getState().flat();
+      const nonEmptyCells = flatState.filter((cell) => cell !== 0);
+
+      expect(nonEmptyCells).toHaveLength(2);
+
+      expect(
+        nonEmptyCells.every((cell) => [2, 4].includes(cell)),
+      ).toBe(true);
```
mgrinko, line 114: You don't need `.every` for 2 cells

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+
+    it('should generate the board with 4s less often than with 2s', () => {
+      const game2048 = new Game();
+
+      let statesCombined = [];
+      let twos = 0;
+      let fours = 0;
+
+      for (let i = 0; i < 100; i++) {
+        game2048.start();
```
mgrinko, line 125: have we already tested that the second `start` call clears the prev state and creates a new one?

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+
+      statesCombined.forEach((cell) => {
+        if (cell === 2) {
+          twos++;
+        } else if (cell === 4) {
+          fours++;
+        }
+      });
+
+      expect(fours < twos).toBe(true);
```
mgrinko, line 143: It you already tested that only 2 and 4 can appear 
you can just calculate the sum and check if it is `< 300`

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+              } else {
+                cellPositions.set(key, 1);
+              }
+            }
+          });
+        });
+      }
+
+      expect(cellPositions.size >= 8).toBe(true);
+      expect(cellPositions.size <= 16).toBe(true);
```
mgrinko, line 172: Random test can fail randomly so in some rear cases it will fail without a reason

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+              const key = `${rowIndex}${cellIndex}`;
+
+              if (cellPositions.has(key)) {
+                cellPositions.set(key, cellPositions.get(key) + 1);
+              } else {
+                cellPositions.set(key, 1);
+              }
+            }
+          });
+        });
```
mgrinko, line 168: Why don't you just use Set to count indexes in a flatState?
becase you don't really use values

tests/Game.class.test.js
mgrinko, line 172: You don't really need to check that it is `<= 16` it is true by the definition of state

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+        game2048.start();
+        game2048.moveLeft();
+
+        const state = game2048.getState();
+
+        state.forEach((row, rowIndex) => {
+          const cursor = initialState[rowIndex].find((cell) => cell !== 0);
+
+          expect(checkIsCorrectlyAligned(row, 'start', cursor)).toBe(true);
+        });
```
mgrinko, line 217: just check the first element in each line 
you don't need `checkIsCorrectlyAligned`

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+            const newCells = row.slice(countBeforeMove).filter((cell) => cell !== 0);
+
+            return [
+              ...acc,
+              ...newCells,
+            ];
+          }, []);
+
+        expect(addedCells.length).toBe(1);
+        expect([2, 4].includes(addedCells[0])).toBe(true);
```
mgrinko, line 255: You don't need all this, because you definitelly know that a new 2 of for will appear in the 3rd column 

So you can just check that 3rd column has 1 non empty cell and is it 2 or 4

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+
+          expect(checkIsCorrectlyAligned(row, 'start', cursor)).toBe(true);
+
+          nonEmptyCells.forEach((cell) => {
+            expect(row.includes(cell)).toBe(true);
+          });
+        });
+      });
+
+      it('should merge cells with the same value', () => {
```
mgrinko, line 286: before testing this we should check that cells are not merged if they are not equal

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+          expect(row.includes(16)).toBe(true);
+        });
+      });
+
+      it('should merge multiple cells with the same value to the left', () => {
+        const initialState = [
+          [8, 8, 16, 16],
+          [16, 16, 8, 8],
+          [32, 32, 8, 8],
+          [16, 16, 16, 16],
```
mgrinko, line 311: this line should go in a separate test 
that 3 element in a row are not merged all together

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+        expect(getCell(2, 3)).toBe(32);
+        expect(getCell(3, 0)).toBe(32);
+        expect(getCell(3, 1)).toBe(16);
+        expect(getCell(3, 2)).toBe(16);
+        expect(getCell(3, 3)).toBe(32);
+      });
+    });
+  });
+
+  describe('game status', () => {
```
mgrinko, line 805: I think we should test all the statuses here (even the simple ones)

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+    it('should be a "lose" when no more moves are possible', () => {
+      const game2048 = new Game([
+        [128, 128, 0, 8],
+        [16, 8, 16, 32],
+        [8, 16, 32, 64],
+        [16, 32, 0, 128],
+      ]);
+
+      game2048.start();
+      game2048.moveLeft();
```
mgrinko, line 842: I think that 2 new items will appear after the start 
So the game should be in a `lose` status before the `moveLeft`

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+  describe('game score', () => {
+    it('should be updated after cells are merged horizontally', () => {
+      const game2048 = new Game([
+        [16, 0, 32, 0],
+        [32, 0, 8, 8],
+        [0, 8, 8, 0],
+        [8, 8, 0, 16],
+      ]);
+
+      game2048.start();
```
mgrinko, line 857: It would be great to check the score right before the move
To ensure that the status calculation in the reason but not the initial status with a non empty cell

tests/Game.class.test.js
```diff
@@ -0,0 +1,916 @@
...
+      game2048.moveRight();
+
+      expect(game2048.getScore()).toBe(192);
+    });
+  });
+
+  describe('game reset', () => {
+    it('should reset the game score, status and set the board to the empty state after restart', () => {
```
mgrinko, line 893: it would be also great to check if we can reset to a non 0 initial state
As we required it in the description

tests/test.helpers.js
```diff
@@ -0,0 +1,86 @@
...
+        if (cursor !== null && cell === cursor) {
+          break;
+        }
+      }
+
+      return !isZeroFoundEnd || cursor !== null;
+    }
+
+    default:
+      throw new Error('Unknown alignment');
```
mgrinko, line 48: This is a test, you control the code fully
So this case can newer happen!!!

tests/test.helpers.js
```diff
@@ -0,0 +1,86 @@
...
+          return false;
+        }
+
+        if (cursor !== null && cell === cursor) {
+          break;
+        }
+      }
+
+      return !isZeroFoundEnd || cursor !== null;
+    }
```
mgrinko, line 45: Why did you combine this 2 cases if they don't have anything in common?

tests/test.helpers.js
```diff
@@ -0,0 +1,86 @@
+'use strict';
+
+const checkIsCorrectlyAligned = (cellsArray, alignment, cursor = null) => {
```
mgrinko, line 3: I think this function is redundunt
You can check what you need much simpler

mgrinko (changes requested):

tests/Game.class.test.js
sergii-nosachenko: Why using `.every` is bad? Do you propose I use such an option?

```js
expect([2, 4].includes(nonEmptyCells[0])).toBe(true);
expect([2, 4].includes(nonEmptyCells[1])).toBe(true);
```

tests/Game.class.test.js
sergii-nosachenko: I'm not sure if using "sum < 300" is more understandable than just comparing total counts.

tests/Game.class.test.js
sergii-nosachenko: I have increased the number of iterations for the test. However, to be honest, I think it's unnecessary. The test always returns 16 items with a good distribution in my case. Thus, I will set the `cellPositions.size` to be equal to 12 or greater since students should write their code correctly. For a properly implemented random position logic, this test will always be true.

tests/Game.class.test.js
sergii-nosachenko: Using your approach:
1) initial:
```json
[
  [8, 0, 0, 0],
  [0, 16, 0, 0],
  [0, 0, 32, 0],
  [0, 0, 0, 64],
]
```

2) after `start()`:
```json
[
  [8, 0, 0, 0],
  [0, 16, 2, 0],
  [0, 0, 32, 0],
  [0, 0, 4, 64],
]
```

3) student's incorrect implementation in `moveLeft()`:
```json
[
  [8, 0, 0, 0],
  [16, 0, 2, 0],
  [32, 0, 2, 0],
  [4, 0, 0, 64],
]
```

4) test passed, but solution is incorrect

tests/Game.class.test.js
sergii-nosachenko: Why? I have 2 free slots in every row. After `start()` we will get 2 random cells somewhere on the board. I don't know where. After `moveLeft()` we will have such a state, for example:
```json
[
  [8, 2, 16, 0],
  [16, 32, 0, 0],
  [8,32, 0, 0],
  [4, 16, 64, 0],
]
```

So, the game will add a new cell somewhere on the free space, 3rd or 4th column.

My test logic finds that cell based on the two state changes.

tests/Game.class.test.js
sergii-nosachenko: It is done in the previous test called `should stack cells to the left without merging`

tests/Game.class.test.js
sergii-nosachenko: Reasonable, will add it

tests/Game.class.test.js
sergii-nosachenko: There are two empty cells, so after `start()` they will be filled with 2 cells (value 2 or 4). But one move will be possible - to the left. It merges 128 + 128. After this only one cell will be available - 1st row, 4th cell. It will be filled again with 2 or 4. But the board will be blocked in that state, so game is lost.

tests/test.helpers.js
sergii-nosachenko: Described why it is not so simple in the comment before

tests/test.helpers.js
sergii-nosachenko: I like to keep things clear and consistent - the switch-case block needs to have a `default` case, so I throw an error here.

tests/test.helpers.js
sergii-nosachenko: Replaced with 2 separate functions

src/modules/Game.class.js
```diff
@@ -0,0 +1,74 @@
...
+   *  [0, 0, 0, 0],
+   *  [0, 0, 0, 0],
+   *  [0, 0, 0, 0]]
+   *
+   * If passed, the board will be initialized with the provided
+   * initial state.
+   */
+  constructor(
+    initialState,
+  ) {}
```
mgrinko, line 25:
```suggestion
  constructor(initialState) {}
```

tests/Game.class.test.js
mgrinko: ok
missed that move

mgrinko (approved):

PR #678, opened 2024-01-25
2048

HYPER2307:
[DEMO LINK](https://hyper2307.github.io/js_2048_game/)


DarkMistyRoom (changes requested):
Good job! Just one comment: a new item appears even if there are no movements or merges on the field. Fix it, please

SanyaBratashchuk (approved):

PR #679, opened 2024-01-27
simplify tests

mgrinko:
For now I did not touch `moveRight`, `moveUp` and `moveDown` tests
If approved will use the same approach as for `moveLeft`


cypress/integration/js2048Game.spec.js
```diff
@@ -9,7 +9,7 @@ Cypress.Commands.add('shuffleBoxes', (arrow1, arrow2, times) => {
   }
 });
 
-describe('2048 game UI', () => {
+describe.skip('2048 game UI', () => {
```
mgrinko, line 12: They took too long
Will fix them later

PR #682, opened 2024-01-30
add task solution

ddl13:
[DEMO LINK](https://ddl13.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,289 @@
 'use strict';
 
-// write your code here
+let board;
+let score = 0;
+const rows = 4;
+const columns = 4;
+
+const startBtn = document.querySelector('.start');
+const scoreItem = document.querySelector('.game-score');
+const restartBtn = document.querySelector('.restart');
+const againBtn = document.querySelector('.again');
+const containerLose = document.querySelector('.container-lose');
+
+let bestScore = localStorage.getItem('2048-highScore') || 0;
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,289 @@
...
+  }
+  checkForWin();
+}
+
+document.addEventListener('keyup', (e) => {
+  if (e.code === 'ArrowLeft') {
+    slideLeft();
+    setNew();
+  } else if (e.code === 'ArrowRight') {
+    slideRight();
```
etojeDenys, line 99: better to use switch case instead the if/else

src/scripts/main.js
```diff
@@ -1,3 +1,289 @@
...
+      tile.classList.add('tile--2048');
+    }
+  }
+  checkForWin();
+}
+
+document.addEventListener('keyup', (e) => {
+  if (e.code === 'ArrowLeft') {
+    slideLeft();
+    setNew();
```
etojeDenys, line 97: you can move it out of the if and add after all cases only one time

etojeDenys (changes requested):
if we don't have any cells to merge then do not add another one

https://github.com/mate-academy/js_2048_game/assets/70220667/2fa60907-50d8-4ebb-9163-ebfc2cb3d268

l4st1m0za (approved):
Good job!

PR #692, opened 2024-02-04
Develop

Misha408:
[DEMO LINK](https://misha408.github.io/js_2048_game/)


DarkMistyRoom (changes requested):
- a new tile shouldn't appear when there are no movements on the board
- please fix the logic of merging. Here is an example. I've written below the previous state. Then I press left and here is what we get. In this case 4 should be on the 2nd cell from the left
image

- hide the `lose` message after restart. And start a new game immediately after pressing `restart`
image

- clean the score after starting a new game
- a user should win after reaching 2048 number on a tile, not on the score
image

PR #694, opened 2024-02-05
2048

Nadinnnka:
- [DEMO LINK](https://Nadinnnka.github.io/js_2048_game/)


SanyaBratashchuk (changes requested):
New cell should not appear after press key if there is no available move in this direction

Andriy-Shymkiv (approved):
good job!

PR #697, opened 2024-02-09
add task solution

DariaSydorenko:
Я не змогла запустити тести, тому орієнтувалась тільки на те, щоб гра працювала в браузері(

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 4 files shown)


IvanFesenko (changes requested):
Looks good, Check my comments:
- New cells should not generate in case no cell move were done
- After lose and restart - game stop working

maxim2310 (changes requested):
still have this bug 
- New cells should not generate in case no cell move were done
on screen if i click right new cell must not appear but it does
image
image

loralevitska (changes requested):
After changes the game crashed

https://github.com/mate-academy/js_2048_game/assets/70009205/f8ca0411-c9d4-4881-9472-2b1bf46b9a4e

DariaSydorenko:
That problem was after the previous commit. I've done 'npm run deploy' and now it should work

SanyaBratashchuk (approved):

PR #699, opened 2024-02-11
2048

FoReWwEr:
https://FoReWwEr.github.io/js_2048_game/


maxim2310 (changes requested):

1. it must be one button
image
2. New cells should not generate in case no cell move were done.
now if i click right will be create new cell but its wrong behavior 
image
image
3. this benner must disappear after start game 
image

PR #704, opened 2024-02-14
game was created

BoGoBooM:
[DEMO LINK](https://bogoboom.github.io/js_2048_game/)


maxim2310 (changes requested):
Almost there) Just one more thing needs to be fixed

need to show banner that inform about lose game
image

maxim2310 (changes requested):
Still not have it
image

vadiimvooo (approved):
LGTM

PR #705, opened 2024-02-14
Solution

Zava96:


maxim2310 (changes requested):
Yor demo link is not working, please fix it and make request review again

Zava96:
> Yor demo link is not working, please fix it and make request review again

It works!

maxim2310 (changes requested):
Almost there) Just one more thing needs to be fixed
after i restart game banner still showing 
image

src/scripts/main.js
```diff
@@ -1,7 +1,246 @@
...
+    if (type.code === 'ArrowUp') {
+      slideUp();
+      setNewCell();
+    }
+
+    if (type.code === 'ArrowDown') {
+      slideDown();
+      setNewCell();
+    }
+  }
```
etojeDenys, line 65: better to just use switch/case instead fo the multiple if/else

src/scripts/main.js
```diff
@@ -1,7 +1,246 @@
...
+
+  startBtn.classList.add('restart');
+  startBtn.innerText = 'Restart';
+});
+
+document.addEventListener('keyup', (type) => {
+  if (startBtn.classList.contains('restart')) {
+    if (type.code === 'ArrowLeft') {
+      slideLeft();
+      setNewCell();
```
etojeDenys, line 48: you can write it only one time after all cases

etojeDenys (changes requested):
image
image

loose message doesn't appear (

maxim2310 (approved):
Well done!!!

PR #710, opened 2024-02-20
Develop

Reaffith:
[DEMO LINK](https://Reaffith.github.io/js_2048_game/)


maxim2310 (changes requested):
1. would be nice to make cells square 
2. new cells must not appear if we havent any move or merge, now you can click just right for chec this bug

l4st1m0za (approved):
Good job!

PR #714, opened 2024-02-21
beta

Stan1slav123:
- [DEMO LINK](https://Stan1slav123.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,7 +11,7 @@
       <div class="game-header">
         <h1>2048</h1>
         <div class="controls">
-          <p class="info">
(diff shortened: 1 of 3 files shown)


maxim2310 (changes requested):

1. i just clicked on reset couple time, it shouldn't be like this 
image
 2.after i loose game i click restart and appeared only one block
image
 3. find bug in merge logic, look:
 here we have four blocks of 2 on bottom line and after i click right i expect that they will merge and will be like two blocks of 4, but look what happened
image
image

PR #716, opened 2024-02-22
solutionV1

kapesha:
solution

I really dont understand, what should I get from some methods, so i would be gratefull for advice


l4st1m0za (approved):
Good job!

PR #721, opened 2024-02-24
add task solution"2048Game"

Andriy-Kutyshenko:
https://Andriy-Kutyshenko.github.io/js_2048_game/


etojeDenys (changes requested):
your demo link doesn't work, try to fix that and then re-request me

lerastarynets (changes requested):
All good, but your game starts by default when the page is loaded, but it should on;y start on Start button click

TarasHoliuk (approved):
With animations :fire: 
Well done :heavy_check_mark:

PR #723, opened 2024-02-26
testSolution

StageUa:
- [DEMO LINK](https://stageua.github.io/js_2048_game/)


src/modules/Game.class.js
```diff
@@ -1,68 +0,0 @@
-'use strict';
```
TarasHoliuk, line 1: why have you moved this class to `main.js`
That's the reason tests can't run

Please move it back to a separate file and do the following:


// Uncomment the next lines to use your game instance in the browser
// const Game = require('../modules/Game.class');
// const game = new Game();

src/scripts/main.js
```diff
@@ -1,7 +1,272 @@
...
+  }
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+  moveLeft() {
+    let moved = false;
 
-// Write your code here
+    for (let i = 0; i < 4; i++) {
```
TarasHoliuk, line 17: Optional: The logic you have written in every `move` method is similar. have you tried to re-use this logic? If not - try to do it

TarasHoliuk (changes requested):
Everything works as expected, but left 2 comments about code

P.S. Provide demo link in the PR description

src/modules/Game.class.js
StageUa: const Game = require('../modules/Game.class');
const game = new Game();

require doesn`t work in browser. Browser send error: main.js:4 Uncaught ReferenceError: require is not defined

src/scripts/main.js
etojeDenys: image
try to fix it, it will simplify the logic

etojeDenys (changes requested):
1. loose message doesn't appear
image
image

StageUa:
> Still not fixed from the previous review image
> 
> Also, will be better if you add favicon on the page image

PR #729, opened 2024-02-29
game 2048

reznik-denis:
- [DEMO LINK](https://reznik-denis.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,7 +1,110 @@
...
+      countKeyPress++;
+    } else if (e.code === 'ArrowUp') {
+      game.moveUp();
+      updateElement();
+      countKeyPress++;
+    } else if (e.code === 'ArrowDown') {
+      game.moveDown();
+      updateElement();
+      countKeyPress++;
+    }
```
etojeDenys, line 91: better to use switch/case here instead of multiple if/else

src/scripts/main.js
```diff
@@ -1,7 +1,110 @@
...
+document.addEventListener('keyup', (e) => {
+  if (game.getStatus() === 'playing') {
+    if (e.code === 'ArrowLeft') {
+      game.moveLeft();
+      updateElement();
+      countKeyPress++;
+    } else if (e.code === 'ArrowRight') {
+      game.moveRight();
+      updateElement();
+      countKeyPress++;
```
etojeDenys, line 82: you can write it only once after all cases

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #731, opened 2024-03-02
add task solution   - [DEMO LINK](https://Chumakv.github.io/js_2048_game/)

Chumakv:
- [DEMO LINK](https://Chumakv.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,438 @@
 'use strict';
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const buttonStart = document.querySelector('.button');
+const messageStart = document.querySelector('.message-start');
+const messageWin = document.querySelector('.message-win');
+const messageLose = document.querySelector('.message-lose');
+const cells = document.querySelectorAll('.field-cell');
+const score = document.querySelector('.game-score');
+const addScore = document.querySelector('.add-points');
 
-// Write your code here
+const twoOrFour = () => {
+  let value;
+
+  if (Math.random() <= 0.1) {
+    value = 4;
+  } else {
+    value = 2;
+  }
+
+  return value;
+};
+
+const randomNewCell = () => {
+  let rando;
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,438 @@
...
+  do {
+    rando = Math.floor(Math.random() * cells.length);
+  } while (cells[rando].textContent !== '');
+
+  return rando;
+};
+
+const cellStyle = () => {
+  cells.forEach(cell => {
+    cell.classList.remove(
```
TarasHoliuk, line 35: easier just to set the default (`field-cell`) class to all cells

src/scripts/main.js
```diff
@@ -1,7 +1,438 @@
...
+
+    switch (cell.textContent) {
+      case '':
+        break;
+      case '2':
+        cell.classList.add('field-cell--2');
+        break;
+      case '4':
+        cell.classList.add('field-cell--4');
+        break;
```
TarasHoliuk, line 50:
```suggestion
      case '2':
      case '4':
        cell.classList.add(`field-cell--${cell.textContend}`);
        break;
```
you may join all cases (except 2048) like this

src/scripts/main.js
```diff
@@ -1,7 +1,438 @@
...
+    && notEmpty === true
+  ) {
+    let changesOccurred = false;
+
+    for (let i = 0; i < cells.length; i++) {
+      let target = i;
+
+      const mover = () => {
+        if (
+          target > 3
```
TarasHoliuk, line 160: quite difficult to understand logic (a lot of unclear conditions in `if`) - try to make it simpler (at least save conditions result in variables with meaningful names)

Also, maybe you can reuse part of the logic inside each arrow button (action.key === 'ArrowDown') handler

TarasHoliuk (changes requested):
Works good, nice animations :fire: 

Added a few comments to improve your code

anastasiiavorobiova (approved):
Great work!

PR #732, opened 2024-03-04
sol

ANTI-bogdan:
[DEMO LINK](https://ANTI-bogdan.github.io/js_2048_game/)


anastasiia-tilikina (changes requested):
looks like something went wrong((
your demo doesn't work(

PR #742, opened 2024-03-07
Develop

MarekSax:
[DEMO](https://mareksax.github.io/js_2048_game/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ dist
 raw_reports
 cypress/screenshots
 cypress/videos
-reports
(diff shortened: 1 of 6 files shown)


yevhenii-pyl (approved):
Great job: this is well-structured and readable code. 🥇 

The only thing I would have changed here are statuses. Just to make it cleaner and avoid possible type, it's always preferable to create an object for statuses, like:

```
const GAME_STATUSES = Object.freeze( {
  playing: 'playing'
})
```

This way you will make sure that this object will never be modified somewhere in your codebase and always have a single source of truth for statuses. 


Other than that - you definitely put your time and effort into the assignment, once again, - good job!

PR #746, opened 2024-03-10
all

Stan1slav123:
- [DEMO LINK](https://Stan1slav123.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,307 @@
-'use strict';
+"use strict";
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+let score = 0;
+let gameWon = false;
+let gameOver = false;
+const board = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
 
-// Write your code here
+function initializeBoard() {
+  addRandomTile();
+  addRandomTile();
+  score = 0;
+  gameWon = false;
+  gameOver = false;
+  updateBoard();
+  document.querySelector(".message-start").classList.add("hidden");
+  document.querySelector(".button.start").textContent = "Restart";
+  document.querySelector(".button.start").classList.add("restart");
+}
+
+function addRandomTile() {
+  const emptyCells = [];
+
+  for (let r = 0; r < 4; r++) {
+    for (let c = 0; c < 4; c++) {
+      if (board[r][c] === 0) {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,307 @@
...
+}
+
+function move(direction) {
+  if (gameWon || gameOver) {
+    return;
+  }
+
+  let moved = false;
+
+  if (direction === "up") {
```
TarasHoliuk, line 78: you can use `switch` here

src/scripts/main.js
```diff
@@ -1,7 +1,307 @@
...
+      }
+    }
+  }
+
+  return true;
+}
+
+document.addEventListener("keydown", function (typeOfEvent) {
+  if (typeOfEvent.key === "ArrowUp") {
+    move("up");
```
TarasHoliuk, line 255: just pass `typeOfEvent.key` to move, then handle strings like `ArrowUp` there (instead of `up`)

src/scripts/main.js
```diff
@@ -1,7 +1,307 @@
...
+  for (let r = 0; r < 4; r++) {
+    for (let c = 0; c < 4; c++) {
+      if (board[r][c] !== 0) {
+        flag = true;
+        break;
+      }
+    }
+  }
+
+  if (flag === true) {
```
TarasHoliuk, line 277: what is `flag`? rename to make it more clear (for example `isGameStarted`, etc.)

src/scripts/main.js
```diff
@@ -1,7 +1,307 @@
...
+          if (board[nextRow][col] === 0) {
+            board[nextRow][col] = board[row][col];
+            board[row][col] = 0;
+            row = nextRow;
+            moved = true;
+          } else if (board[nextRow][col] === board[row][col]) {
+            board[nextRow][col] *= 2;
+            score += board[nextRow][col];
+            board[row][col] = 0;
+            moved = true;
```
TarasHoliuk, line 208: maybe you can make your code reusable. In every `move{Arrow}` function you do similar things
For example, create a function `mergeCells` that accepts correct indices (col, row, etc) and changes `board` array, etc

TarasHoliuk (changes requested):
GJ - everything works as expected :+1: 
Added a few comments to improve your code

anastasiiavorobiova (changes requested):
Great work! Some gaps appeared when some numbers were added. Consider fixing it 
Screenshot 2024-03-10 at 15 10 31

Stan1slav123:
I don' understand the issue, can you give me more information please

anastasiiavorobiova (approved):
Great work!

PR #753, opened 2024-03-18
add task solution

MatiushechkoMark:
[DEMO LINK](https://MatiushechkoMark.github.io/js_2048_game/)


maxim2310 (approved):
Good work, very good animation!!!

just from my POW would be decrease duration of animation

PR #762, opened 2024-03-28
First Commit

rwlp:
Hi everyone! 

This is the first commit, To be more productivity, feel free to analyze only JavaScript code, focus on logic, design patterns, good pratices and conventios of code writing.

I'll treat the others aspects in a second PR.

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -207,7 +207,7 @@ class Game {
     }
   }
 
-  // Atention this is used to revert each line of this.#state, dont to revert any array. 
+  // Revert each line of this.#state, dont use to revert any array. 
   #reverse() {
     this.#state.forEach((e, i, arr) => arr[i].reverse());
   }
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -6,6 +6,13 @@
  * Feel free to add more props and methods if needed.
  */
 class Game {
+
+  // This are variables used in entire instance of class
+  #SIZE = 4;
+  #score = 0;
+  #state = [];
+  #status = '';
+  #initialState = undefined;
   /**
    * Creates a new game instance.
    *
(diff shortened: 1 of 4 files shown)


src/modules/Game.class.js
```diff
@@ -50,19 +138,168 @@ class Game {
...
-   * Starts the game.
+   * Put 2 or 4 in freely cells if they exists 
    */
-  start() {}
+  #randomizer() {
+    const empties = [];
+    let idxSorted;
+
+    for (let line = 0; line < 4; line++) {
+      for (let column = 0; column < 4; column++) {
```
EdPirro, line 182: Value 4 seems to be hard-coded here; did you mean to use `this.#SIZE`?

src/modules/Game.class.js
```diff
@@ -50,19 +138,168 @@ class Game {
...
+    this.#state.forEach((e, i, arr) => arr[i].reverse());
+  }
+
+  // Check if movements isn't possible than change this.status.
+  #checkIfLose() {
+    let allSides = [];
+
+    for (let column = 0; column < this.#SIZE; column++) {
+      for (let line = 0; line < this.#SIZE; line++) {
+        if (this.#state[line][column] !== 0) {
```
EdPirro, line 221: What if it `this.#state[line][column]` is 0; what does it mean? Use early returns whenever possible to avoid unnecessary processes

src/modules/Game.class.js
```diff
@@ -50,19 +138,168 @@ class Game {
...
-   * Resets the game.
+   * This method copy two dimensional array.
+   * In this code is always used to copy the state Two dimensional Array
+   *
+   * @param {Array} arr
+   * @returns
    */
-  restart() {}
+  #getStateCopy(arr = this.#state) {
+    return arr.map((e) => e.slice());
```
EdPirro, line 258: `.slice` is not meant for array copy; use spreading instead `[...e]`

src/modules/Game.class.js
```diff
@@ -50,19 +138,168 @@ class Game {
...
+  }
+
+  // Make left movement
+  #makeMove() {
+    let limit;
 
-  // Add your own methods here
+    for (let line = 0; line < 4; line++) {
```
EdPirro, line 265: 4 -> `this.#SIZE`

src/modules/Game.class.js
```diff
@@ -50,19 +138,168 @@ class Game {
...
+        } else if (this.#state[line][column + 1] === 0) {
+          for (let idx = column + 1; idx < this.#SIZE - 1; idx++) {
+            this.#state[line][idx] = this.#state[line][idx + 1];
+            this.#state[line][idx + 1] = 0;
+          }
+
+          limit--;
+          column--;
+
+          continue;
```
EdPirro, line 288: These first two ifs are used to ignore 0's why not `.filter` it and then refill missing elements with 0?

src/modules/Game.class.js
```diff
@@ -50,19 +138,168 @@ class Game {
...
+        } else if (
+          this.#state[line][column] === this.#state[line][column + 1]
+        ) {
+          this.#state[line][column] *= 2;
+          this.#state[line][column + 1] = 0;
```
EdPirro, line 293: Instead of adding a 0 that will be removed in the next iteration handle it here

src/scripts/main.js
```diff
@@ -1,7 +1,98 @@
...
+    }
+    setStyles();
+    score.textContent = game.getScore();
+  },
+  win: () => {
+    actions.playing();
+    messages[0].className = 'message message-lose hidden';
+    messages[1].className = 'message message-win';
+    messages[2].className = 'message message-start hidden';
+    messages[3].className = 'message message-restart hidden';
```
EdPirro, line 52: See [element.classList](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) and simply remove or add `hidden` as needed instead of redefining the element's className

src/scripts/main.js
```diff
@@ -1,7 +1,98 @@
...
+  if (game.getStatus() === 'playing') {
+    if (eventKey.key === 'ArrowDown') {
+      game.moveDown();
+
+      bindState(game.getStatus());
+    } else if (eventKey.key === 'ArrowUp') {
+      game.moveUp();
+      bindState(game.getStatus());
+    } else if (eventKey.key === 'ArrowRight') {
+      game.moveRight();
```
EdPirro, line 85: Not updating game state after `game.moveRight()`

EdPirro (changes requested):
Great effort, @rwlp! Looking forward to seeing how this project turns out 🤩! Since I couldn't yet test it, I simply added a few suggestions; please review them and ask me if you have any questions!

src/modules/Game.class.js
```diff
@@ -21,27 +28,108 @@ class Game {
...
+  moveLeft() {
+    if (this.#status === 'playing') {
+      const previousState = this.#getStateCopy().flat();
+
+      this.#makeMove();
+
+      if (this.#state.flat().some((e, i) => e !== previousState[i])) {
+        this.#randomizer();
+        this.#checkIfLose();
+      }
```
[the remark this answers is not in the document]
rwlp: I think that is not possible, because each movement make some operations( reverse, tranpose etc. ).
For example the left movement just call makeMove() method but the right movement use the sequence of methods reverse()-> makeMove() -> reverse(). Each movement have your own sequence of methods always reusing code in this methods(reverse, makeMove, tranpose);

src/modules/Game.class.js
rwlp: The logic is basically this steps you starts in the first position: 

'0 0 2 2' you is zero then ->  '0 2 2 0' you is zero then -> you is not zero then '2 2 0 0'; -> '4 0 0 0' ( see the limit var is dinamically changed trying to avoid more calls if not necessary). Is possible many combinations '2 0 2 0' '0 4 0 2' etc. 
I don't test but the intention of this is that code is able to NxN sizes not only 4x4.

I think a good improvement change the for limit on lines 280 and 270. Is that make sense?

src/modules/Game.class.js
rwlp: This zero is because I need to change a not zero value to zero because I have a sum. The case if this zero need to removed( acctually replaced) or not is checked in next step of loop. Can you clarify if is necessary additional code?

rwlp:
[DemoLink](https://rwlp.github.io/js_2048_game/)

DarkMistyRoom (changes requested):
Great job, but it seems like the right arrow doesn't work

rwlp:
> Great job, but it seems like the right arrow doesn't work

Thank you git-hub invasor!

src/modules/Game.class.js
```diff
@@ -50,19 +99,170 @@ class Game {
...
+  #checkIfLose() {
+    let allSides = [];
+
+    for (let column = 0; column < this.#SIZE; column++) {
+      for (let line = 0; line < this.#SIZE; line++) {
+        if (this.#state[line][column] !== 0) {
+          allSides.push(
+            this.#state[line - 1] !== undefined
+              ? this.#state[line - 1][column]
+              : undefined,
```
EdPirro, line 188: May use the [Optional Chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) instead of ternary `this.#state[line - 1]?.[column]`

EdPirro (approved):
Impressive job, @rwlp! 👏👏 Your code looks great; very clean and straightforward. Keep it up! 🚀

PR #768, opened 2024-04-07
Develop

IvanBerdnyk:
- [DEMO LINK](https://IvanBerdnyk.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
1. fix the styles here
image


2. score doesn't work
3. start text below should disappear after the game starts
image

4. when I press the up key - nothing happens
image

PR #770, opened 2024-04-11
Solution

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/js_2048_game/)


yevhenii-pyl (approved):
Solid! 
Played with code locally: well structured and detailed. 
However, some places, like `moveTiles` method, may benefit from comments, just to make it cleaner in case you'll have to get back to it after some decent amount of time :)

PR #773, opened 2024-04-15
Implement 2048 game

olafchuszno:
[DEMO LINK](https://olafchuszno.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,6 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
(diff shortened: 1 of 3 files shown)


yevhenii-pyl (approved):
Solid: well structured and easy to follow.  🥇 

However, this code may benefit from comments. You can use this assignment as a good practise on how to comment your code when 'it all makes sense', but still the logic is pretty complicated and it may be hard to follow after some time.

olafchuszno:
I always comment! I noticed that on every other task I've done I was asked to remove them, so I just assumed the same here and removed them before pushing.

PR #778, opened 2024-04-26
2048 game implemented

artemvlasiuk:
[DEMO LINK](https://artemvlasiuk.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,277 @@
 'use strict';
 
-// write your code here
+let board;
+let score = 0;
+const rows = 4;
+const columns = 4;
+
+const startButton = document.querySelector('.start');
+const messageStart = document.querySelector('.message-start');
+
+startButton.addEventListener('click', () => {
+  setGame();
+  setTwo();
+  setTwo();
+  messageStart.classList.add('hidden');
+});
+
+window.onload = function() {
+  setGame();
+};
+
+function setGame() {
+  const gameField = document.getElementsByClassName('game-field')[0];
+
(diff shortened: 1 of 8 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Fix the right padding for this button
image

2. Add a favicon on the page
image

src/scripts/main.js
```diff
@@ -1,3 +1,277 @@
...
+  messageGameOver.classList.remove('hidden');
+
+  return true;
+}
+
+document.addEventListener('keyup', (e) => {
+  switch (e.code) {
+    case 'ArrowLeft':
+      slideLeft();
+      setTwo();
```
etojeDenys, line 145: you can move it after the switch/case and call only once for all cases

etojeDenys (changes requested):
show the restart button, if the game has already started
image

2. if we don't merge any cells then do not create another one

https://github.com/mate-academy/js_2048_game/assets/70220667/e8a442b3-f596-453e-9f7c-18a3373bab32

src/scripts/main.js
artemvlasiuk: > 2. if we don't merge any cells then do not create another one

 - then there will be a problem where it will be impossible to continue the game.

https://github.com/mate-academy/js_2048_game/assets/80981782/428f24ac-d0b3-499c-b1a8-1b8945c47e7d

volodymyr-soltys97 (approved):
Nice work 🔥

PR #779, opened 2024-04-27
solution

MykhailoLoniak:
[DEMO LINK](https://MykhailoLoniak.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,7 +1,338 @@
...
+  messageLose.classList.add('hidden');
+  messageWin.classList.add('hidden');
+  messageStart.classList.add('hidden');
+  document.addEventListener('keydown', keyboardHandler);
+
+  generate();
+  generate();
+};
+
+function keyboard(event) {
```
l4st1m0za, line 71: Be careful with naming of the elements as 'event' is already a global variable

src/scripts/main.js
```diff
@@ -1,7 +1,338 @@
...
+  if (event.key === 'ArrowUp') {
+    keyUp();
+    console.log('Натиснута клавіша "Вгору"');
```
l4st1m0za, line 74: It's better to reduce the console logs

src/scripts/main.js
```diff
@@ -1,7 +1,338 @@
...
+function moveUp() {
+  for (let i = 0; i < 4; i++) {
+    const index = [i, i + width, i + width * 2, i + width * 3];
+    const column = [index.map((e) => cellsArray[e].textContent)];
+    const filteredColumn = column[0].filter((num) => num);
+    const missing = 4 - filteredColumn.length;
+    const zeros = Array(missing).fill(0);
+    const newColumn = filteredColumn.concat(zeros);
+
+    index.forEach((e, index) => {
```
l4st1m0za, line 144:
```suggestion
    index.forEach((e, i) => {
```

The same thing here with index in the forEach. if was declared before

l4st1m0za (approved):
Good job!

PR #784, opened 2024-05-08
add task solution + undo button

DamianAdamczewski:
[DEMO LINK](https://DamianAdamczewski.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,68 +1,483 @@
 'use strict';
 
-/**
- * This class represents the game.
- * Now it has a basic structure, that is needed for testing.
- * Feel free to add more props and methods if needed.
- */
 class Game {
-  /**
-   * Creates a new game instance.
-   *
-   * @param {number[][]} initialState
-   * The initial state of the board.
-   * @default
-   * [[0, 0, 0, 0],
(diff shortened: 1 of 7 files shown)


src/modules/Game.class.js
```diff
@@ -1,68 +1,483 @@
...
+    this.createRandomTile();
+    this.printTiles();
+  }
+
+  restart() {
+    this.clearTheBoard();
+    this.printTiles();
+    this.status = Game.Status.idle;
+    this.score = 0;
+    this.undoStack = []; // Clear undo stack on restart
```
yevhenii-pyl, line 48: There is a small bug here, steps to reproduce:

1. Start the game. 
2. Make a few moves to receive some score. 
3. Click on a 'Restart' button. Note the 'score' value.
4. Click on a 'Start' button. Note the 'score' value. 
5. Make a move. Note a 'score' value. 

Expected behaviour: 
the score equals to 0 as soon as a user clicked on a 'restart' button. 

Current behaviour:
the score is reseted only after step 5.

yevhenii-pyl (approved):
Solid!
Code is clean and covered with meaningful comments, good job!

There is a small bug however, but can't see it as a reason to block the PR. Just check how to fix it on a spare moment.

PR #788, opened 2024-05-15
2048

Ihor-Prodan:
[DEMO LINK](https://Ihor-Prodan.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,413 @@
-'use strict';
+const WINNING_VALUE = 2048;
+const GRID_SIZE = 4;
+const CELL_SIZE = 75;
+const CELL_GAP = 5;
 
-// Uncomment the next lines to use your game instance in the browser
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,413 @@
...
+    return;
+  }
+
+  switch (e.key) {
+    case 'ArrowUp':
+      if (!canMoveUp()) {
+        setupInput();
+
+        return;
+      }
```
etojeDenys, line 220: you can move this logic out of the switch/case and write only one time

etojeDenys (changes requested):

vadiimvooo (changes requested):
Seems like your deployment link unreachable. Please fix it.

vadiimvooo (changes requested):
.

vadiimvooo (approved):
LGTM

PR #804, opened 2024-06-17
2048 game

vintwp:
https://vintwp.github.io/js_2048_game/

Test is not passed. Terminal shows infinite loader
image

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -3,12 +3,6 @@
 document.addEventListener('DOMContentLoaded', () => {
   const Game = require('../modules/Game.class');
   const game = new Game();
-  // const game = new Game([
-  //   [2, 0, 0, 0],
-  //   [0, 4, 0, 0],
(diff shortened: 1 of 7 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,159 @@
...
+
+        break;
+      case 'ArrowRight':
+        game.moveRight();
+        refreshGameField();
+
+        if (gStatus() === 'playing') {
+          changeButtonStartToRestart();
+        }
+
```
etojeDenys, line 112: you can move it out of the switch case, and write only one time

etojeDenys (changes requested):
we don't need to create the cell in this case(we didn't merge any cell)

https://github.com/mate-academy/js_2048_game/assets/70220667/ef1c8f29-b0ec-4141-beba-04be4b05de79

volodymyr-soltys97 (approved):
Good job 👍

PR #806, opened 2024-06-18
problem with moves

pakhomovalex:
[DEMO LINK](https://pakhomovalex.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,38 @@
 'use strict';
 
 // Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
+const game = new Game();
 
 // Write your code here
+const startBtn = document.querySelector('.start');
+const restartBtn = document.querySelector('.restart');
+const arrowsToggle = (event) => {
+  if (event.key === 'ArrowUp') {
+    game.moveUp();
+  } else if (event.key === 'ArrowLeft') {
+    game.moveLeft();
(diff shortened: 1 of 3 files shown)


VitaliyBondarenko1982 (changes requested):
- fix linter errors
Screenshot 2024-06-18 at 17 37 41
- please, check game functionality.  cells not moves

src/scripts/main.js
```diff
@@ -1,7 +1,49 @@
...
+    score.textContent = game.score;
+  } else if (event.key === 'ArrowLeft') {
+    game.moveLeft();
+    score.textContent = game.score;
+  } else if (event.key === 'ArrowRight') {
+    game.moveRight();
+    score.textContent = game.score;
+  } else if (event.key === 'ArrowDown') {
+    game.moveDown();
+    score.textContent = game.score;
```
etojeDenys, line 26: better to use swith/case here instead of multiple if/else

src/scripts/main.js
```diff
@@ -1,7 +1,49 @@
...
+const startBtn = document.querySelector('.start');
+const restartBtn = document.querySelector('.restart');
+
+const score = document.querySelector('.game-score');
+
+// eslint-disable-next-line no-shadow
+const arrowsToggle = (event) => {
+  if (event.key === 'ArrowUp') {
+    game.moveUp();
+    score.textContent = game.score;
```
etojeDenys, line 17: you write it only one time after all cases

etojeDenys (changes requested):

etojeDenys (approved):
🔥

PR #812, opened 2024-06-22
asd

jdifek:
- [DEMO LINK](https://jdifek.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -243,7 +243,10 @@ class Game {
           return true;
         }
 
-        if (i < this.state.length - 1 && this.state[i][j] === this.state[i + 1][j]) {
+        if (
+          i < this.state.length - 1 &&
+          this.state[i][j] === this.state[i + 1][j]
+        ) {
           return true;
         }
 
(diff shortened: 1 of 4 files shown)


GUSILLUS (changes requested):
GJ!

Try to implement game over message, because i don't have available moves but I still can't play

TarasHoliuk:
1. When I lose the game and press `Restart`, then start clicking arrow buttons I see this (button still says Start and I see the message below), but I can play:

image

2. You shouldn't add a new tile when after the arrow button click nothing changed:
- Current field:
image

- I pressed left (nothing changed, but the new tile was added):
image

src/modules/Game.class.js
```diff
@@ -1,68 +1,270 @@
...
+
+    this.addRandomTile();
+
+    if (!this.movesAvailable()) {
+      this.status = Game.STATUS_LOSE;
+    }
+  }
+
+  moveDown() {
+    if (this.status === Game.STATUS_WIN || this.status === Game.STATUS_LOSE) {
```
TarasHoliuk, line 149: Maybe you shouldn't do anything when the status is IDLE as well. See the error in the comments

src/scripts/main.js
```diff
@@ -1,7 +1,84 @@
...
+  } else {
+    button.textContent = 'Start';
+
+    game.restart();
+    updateGame();
+  }
+});
+
+function handleKeyPress(e) {
+  if (e.key === 'ArrowDown') {
```
TarasHoliuk, line 30: You can use switch here

TarasHoliuk (changes requested):
For some reason, tests can't start
Try to fix it (ask for the help in the chat if you need)

PR #815, opened 2024-06-24
Develop

lmuias:
[DEMO LINK](https://lmuias.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -6,63 +6,376 @@
  * Feel free to add more props and methods if needed.
  */
 class Game {
-  /**
-   * Creates a new game instance.
-   *
-   * @param {number[][]} initialState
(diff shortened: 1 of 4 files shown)


src/modules/Game.class.js
```diff
@@ -6,63 +6,376 @@
...
+    this.status = Game.STATUS;
+    this.initialState = initialState;
+    this.state = initialState.map((row) => [...row]);
   }
 
-  moveLeft() {}
-  moveRight() {}
-  moveUp() {}
-  moveDown() {}
+  moveLeft() {
```
etojeDenys, line 30: actually you can try to simplify this logic cause 
moveLeft is just 

```
moveRight()
reverse()
```


we can do the same with move up and down

etojeDenys (approved):
good job

PR #823, opened 2024-07-03
2048 testing solution

mikezhylka:
Hello! I wrote this game but didn't check test requirements and didn't implement getState() and getStatus() methods inside class. If code works, in which I am not sure about because some tests fail, do I still need to implement these methods? 

- [DEMO LINK](https://mikezhylka.github.io/js_2048_game/)


anastasiiavorobiova (changes requested):
Great start! The game doesn't work, it should add just one number per move. It's ok if your methods are named differently

SanyaBratashchuk (approved):

PR #831, opened 2024-07-08
Develop

Alenalenk:
[DEMO LINK](https://alenalenk.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -7,14 +7,7 @@ const startBtn = document.querySelector('.start');
 const messageWin = document.querySelector('.message-win');
 const messageLose = document.querySelector('.message-lose');
 
-const field = [
-  [0, 0, 0, 0],
-  [0, 0, 0, 0],
-  [0, 0, 0, 0],
-  [0, 0, 0, 0],
-];
-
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
Screenshot 2024-07-08 at 14 42 09
better to remove this message in this case

2. left button works kinda strange, it doesn't work immediately
3. and score calculated wrongly
image

src/scripts/main.js
```diff
@@ -1,7 +1,92 @@
...
+
+  if (e.key === 'ArrowUp') {
+    game.moveUp();
+    addField();
+  }
+
+  if (e.key === 'ArrowDown') {
+    game.moveDown();
+    addField();
+  }
```
etojeDenys, line 71: you can simplify it using the switch case and move `addField();` out of cases and write only one time

etojeDenys (changes requested):

anastasiiavorobiova (changes requested):
Great work! Numbers stop move when a user press arrows

anastasiiavorobiova (approved):
Great work!

PR #833, opened 2024-07-08
Solution

kyrylomanko:
[DEMO LINK](https://kyrylomanko.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,94 @@
 'use strict';
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
+const game = new Game();
 
-// Write your code here
+const container = document.querySelector('.container');
+const gameField = document.querySelector('.game-field');
+const button = container.querySelector('.button');
+const gameScore = container.querySelector('.game-score');
+
+const fieldRows = [...gameField.querySelectorAll('.field-row')];
+const fieldCells = fieldRows.map((row) => [...row.children]);
+
+const messages = {
+  idle: container.querySelector('.message-start'),
+  lose: container.querySelector('.message-lose'),
+  win: container.querySelector('.message-win'),
+};
+
+function fillGameField(state) {
+  state.forEach((row, i) => {
+    row.forEach((cell, j) => {
+      const currentCell = fieldCells[i][j];
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -3,92 +3,94 @@
...
-  };
-
-  const action = moveActions[keyEvent.key];
-
-  if (action) {
-    action.call(game);
+// Event Listeners
+document.addEventListener('keydown', (eventKey) => {
+  if (game.getStatus() === 'playing') {
+    if (eventKey.key === 'ArrowDown') {
```
TarasHoliuk, line 75: use `switch` block here instead of multiple `if else` blocks

src/scripts/main.js
```diff
@@ -3,92 +3,94 @@
...
+      game.moveDown();
+      bindState(game.getStatus());
```
TarasHoliuk, line 77: you use `game.getStatus()` a lot of times here - save it into a variable in the start of listener function

src/scripts/main.js
```diff
@@ -3,92 +3,94 @@
...
-const messages = {
-  idle: container.querySelector('.message-start'),
-  lose: container.querySelector('.message-lose'),
-  win: container.querySelector('.message-win'),
+const cells = document.getElementsByClassName('field-cell');
+const score = document.getElementsByClassName('game-score')[0];
+const startAndRestartButton = document.getElementsByClassName('button')[0];
+const messages = document.getElementsByClassName('message');
+
+const bgColorByNumber = {
```
TarasHoliuk, line 11: you can make it a function that accepts number as an argument and returns `field-cell--${number}`. Also, you can check - `if (!number) { return ''; }`

TarasHoliuk (changes requested):
Great job! Added a few small suggestions

TarasHoliuk (changes requested):
Technical request - to sync status with MA platform

Sorry if this message was confusing

TarasHoliuk (changes requested):
Please, add requested by mentor (me 🙂 ) changes and then re-request review

anastasiiavorobiova (changes requested):
Great work! Consider removing the start message 
Screenshot 2024-07-14 at 15 57 34
Also, the fail message should be shown if the game ends

TarasHoliuk (approved):
Well done!

PR #856, opened 2024-07-20
Working 2048 prototype

Teollan:
[DEMO LINK](https://Teollan.github.io/js_2048_game/)

The prototype seems to be working fine. However, some animations wouldn't hurt (would require a quite big redesign of the entire thing, and it isn't a requirement anyway). 

Also I added a debug mode. Press F6 to toggle it (`debug mode: true` should appear in the console). It allows you to manually cycle through cell values by clicking on them to test different combinations (or just cheat and get an easy victory 😄 ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,51 +7,56 @@ In this task, you need to implement the 2048 game like in [this reference](https
 Don't play for too long! We need you to write the code!
 
 Okay, what do we have?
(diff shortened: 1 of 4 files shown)


BudnikOleksii (approved):
Looks great 🚀

PR #864, opened 2024-07-23
add task solution

valost:
https://valost.github.io/js_2048_game/


etojeDenys (approved):
good job, but it's better to add some favicon

PR #867, opened 2024-07-25
add task solution

ivankovbohdan:
[DEMO LINK](https://ivankovbohdan.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,284 @@
 'use strict';
 
-// write your code here
+const gameScore = document.querySelector('.game-score');
+const startButton = document.querySelector('.start');
+const fieldCells = document.querySelectorAll('.field-cell');
+
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+const messageStart = document.querySelector('.message-start');
+
+let score = 0;
+let gameField = [];
+
+function startGame() {
+  score = 0;
+  startButton.textContent = 'Restart';
+  startButton.classList.remove('start');
+  startButton.classList.add('restart');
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,284 @@
...
+
+startButton.addEventListener('click', () => {
+  startGame();
+});
+
+document.addEventListener('keydown', (e) => {
+  switch (e.key) {
+    case 'ArrowUp':
+      moveUp();
+      generateNewNumber();
```
etojeDenys, line 259: you can write it only one time after all cases

src/scripts/main.js
```diff
@@ -1,3 +1,284 @@
...
+          gameField[row][currentCol] = 0;
+        }
+
+        // fieldCells[row + (currentCol - 1) * 4].classList.add(
+        //   `field-cell--${gameField[row][currentCol - 1]}`,
+        // );
+
+        // fieldCells[row + (currentCol) * 4].classList.remove(
+        //   `field-cell--${gameField[row][currentCol]}`,
+        // );
```
etojeDenys, line 213: remove all cases

etojeDenys (changes requested):
don't create a new cell in this case

https://github.com/user-attachments/assets/bbbfed1c-a1d5-4755-90a4-0fe2284ba96f

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The situation with the number 2 is interesting, it has a different background color everywhere, this should be fixed 
image

2. Add a favicon on the page
image

volodymyr-soltys97 (approved):
Well done! 🔥

PR #871, opened 2024-07-29
Develop

HunchakAndrii:
- [DEMO LINK](https://HunchakAndrii.github.io/js_2048_game/)


src/modules/fullCell.class.js
```diff
@@ -0,0 +1,41 @@
+export class fullCell {
```
TarasHoliuk, line 1:
```suggestion
export class FullCell {
```
and filename in PascalCase

src/modules/Game.class.js
```diff
@@ -5,40 +5,189 @@
...
+    this.cellsGroupedByColumn = this.groupCellsByColumn();
+
+    this.cellsGroupedByReversedColumn = this.cellsGroupedByColumn.map(
+      (column) => [...column].reverse(),
+    );
+    this.cellsGroupedByRow = this.groupCellsByRow();
+
+    this.cellsGroupedByReversedRow = this.cellsGroupedByRow.map(
+      (row) => [...row].reverse(),
+      // eslint-disable-next-line function-paren-newline
```
TarasHoliuk, line 36: do you need it?

TarasHoliuk:
When I click any arrows I see a "Loose error message" even if I haven't started the game yet:
image

TarasHoliuk (changes requested):
Please, try to fix tests

HunchakAndrii:
s
i dont know whats wrong in test

src/scripts/main.js
```diff
@@ -1,7 +1,86 @@
...
+}
+
+// eslint-disable-next-line no-shadow
+function handleInput(event) {
+  switch (event.key) {
+    case 'ArrowLeft':
+      if (game.canMoveLeft()) {
+        game.moveLeft();
+        game.createNewTile();
+        game.checkGameOver();
```
etojeDenys, line 51: you can move this logic outside of the switch case and write it only one time

etojeDenys (changes requested):
why do you always create two new cells, There should be only one

etojeDenys (approved):
lgtm

PR #876, opened 2024-07-31
commit message

asushko25:
- [DEMO LINK](https://asushko25.github.io/js_2048_game/)
I add comments

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


etojeDenys (approved):
good job, but better to remove comments, and also you can try to simplify your logic, because: moveRight is just moveLeft and reverse

moveUp is just matrixTranspose and moveLeft and matrixTranspose back

PR #885, opened 2024-08-06
solution

hsvirina:
https://hsvirina.github.io/js_2048_game/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
+    <script
+      src="scripts/main.js"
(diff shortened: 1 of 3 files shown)


lerastarynets (changes requested):
When I click starts, fields remain empty and I need to click any button before numbers appear in cells, this is incorrect, also nothing happens if user loose
Screenshot at Aug 06 17-58-54

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The game ends in error because it is now possible to move up and down 
image

2. Add a favicon on the page 
image

lerastarynets (approved):

PR #892, opened 2024-08-12
add solution

Valentyn-Radobenko:
[DEMO LINK](https://valentyn-radobenko.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,367 @@
 'use strict';
 
 // Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
+const game = new Game();
 
-// Write your code here
+let filledCells = 0;
+const score = document.querySelector('.game-score');
+const initialState = [
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+  [0, 0, 0, 0],
+];
+
+let state = structuredClone(initialState);
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,367 @@
...
+    game.moveUp();
+  }
+
+  if (e.keyCode === 39) {
+    game.moveRight();
+  }
+
+  if (e.keyCode === 40) {
+    game.moveDown();
+  }
```
etojeDenys, line 49: you can simplify it by using the switch/case

src/scripts/main.js
```diff
@@ -1,7 +1,367 @@
...
+    addRandomNumInCells();
+  }
+
+  if ([...cells].some((cell) => cell.textContent === '2048')) {
+    messageWin.className = 'message message-win';
+  }
+
+  if (filledCells === cells.length && !canGameContinue()) {
+    messageLose.className = 'message message-lose';
+  }
```
etojeDenys, line 366: you can move these lines of code after switch/case
image
 and write it only one time

etojeDenys (changes requested):

src/scripts/main.js
Valentyn-Radobenko:
```
if (isSomethingChanged) {
    addRandomNumInCells();
  }
```
I can't move this line of code after switch/case, because it's a local variable for every move method. It's probably possible, but maybe it's a bit of extra work and it's not worth it. However, I have moved the other lines of code mentioned.

etojeDenys (approved):
good job, but we need to create two cells in this case 
image

PR #897, opened 2024-08-14
add 2048

YuliiaKosenchuk:
- [DEMO LINK](https://YuliiaKosenchuk.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,7 +1,87 @@
...
+const gameOverText = document.querySelector('.game-over-text');
+const game = new Game();
 
-// Write your code here
+document.addEventListener('keydown', (e) => {
+  if (game.getStatus() === 'playing') {
+    switch (e.key) {
+      case 'ArrowLeft':
+        game.moveLeft();
+        updateUI();
```
etojeDenys, line 14: you can write it only one time after the switch case

etojeDenys (changes requested):
small fix

PR #935, opened 2024-09-14
add task solution

Iryna1802:
[DEMO LINK](https://Iryna1802.github.io/js_2048_game/)


TarasHoliuk (approved):
Great job! Approved!

One improvement you might try is to reuse part of moving logic - it's similar in each "move" function. But only if you want - no need to re-request review - already approved 🙂

PR #936, opened 2024-09-14
solution

Kozubowicz:
[DEMO](https://kozubowicz.github.io/js_2048_game/)


natalia-klonowska (changes requested):
you deleted configuration files so you need to add it back so tests could work
image

PR #944, opened 2024-09-20
develop

cyberte4:
[DEMO LINK](https://cyberte4.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -6,63 +6,289 @@
  * Feel free to add more props and methods if needed.
  */
 class Game {
-  /**
-   * Creates a new game instance.
-   *
-   * @param {number[][]} initialState
-   * The initial state of the board.
-   * @default
-   * [[0, 0, 0, 0],
-   *  [0, 0, 0, 0],
-   *  [0, 0, 0, 0],
-   *  [0, 0, 0, 0]]
-   *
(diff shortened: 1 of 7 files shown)


anastasiiavorobiova (changes requested):
Great work! The restart after failing doesn't work. Consider fixing it

cyberte4:
> Great work! The restart after failing doesn't work. Consider fixing it

I know,i need help with it

volodymyr-soltys97 (changes requested):
Good job 👍 
Add a favicon on the page 
image

src/modules/Game.class.js
```diff
@@ -6,63 +6,303 @@
...
+
+    if (randomValue < 0.9) {
+      this.grid[randomRow][randomCol] = 2;
+    } else {
+      this.grid[randomRow][randomCol] = 4;
+    }
+  }
+
+  renderBoard() {
+    const cells = document.querySelectorAll('.field-cell');
```
TarasHoliuk, line 89: That's the reason tests can't run - they run in node.js and there are no `window` or `document` (available only in the browser)

FYI

TarasHoliuk (approved):
Everything works as expected - approved ✔️ 

Added 1 comment about tests - just FYI

PR #945, opened 2024-09-20
Develop

VikaChereushenko:
- [DEMO LINK](https://VikaChereushenko.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -6,9 +6,8 @@ const game = new Game();
 
 // Write your code here
 
-game.start();
-
 const body = document.querySelector('body');
+const button = document.querySelector('.button');
 
 body.addEventListener('keydown', (e) => {
   if (e.key === 'ArrowUp') {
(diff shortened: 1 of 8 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,30 @@
...
+
+body.addEventListener('keydown', (e) => {
+  if (e.key === 'ArrowUp') {
+    game.moveUp();
+  } else if (e.key === 'ArrowDown') {
+    game.moveDown();
+  } else if (e.key === 'ArrowLeft') {
+    game.moveLeft();
+  } else if (e.key === 'ArrowRight') {
+    game.moveRight();
```
etojeDenys, line 28: better to use the swith case here instead of the multiple if/else

etojeDenys (changes requested):
https://github.com/user-attachments/assets/086a28f4-ed6d-45fd-b1c6-87c1009efda0

we don't need to create a cell if we didn't merge any cells after the move

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Add a favicon on the page 
image

2. Need to show a loss message when the game is over
image

etojeDenys (changes requested):
I still should have the ability to move Up or Down
image

VitaliyBondarenko1982 (approved):
Looks good.

PR #946, opened 2024-09-21
add task solution

BohdanDymydiuk:
[DEMO LINK](https://BohdanDymydiuk.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,134 @@
+/* eslint-disable function-paren-newline */
 'use strict';
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
+const game = new Game();
 
-// Write your code here
+const cells = document.querySelectorAll('.field-cell');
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,134 @@
...
+    case '128':
+      element.classList.add('field-cell--128');
+      break;
+    case '256':
+      element.classList.add('field-cell--256');
+      break;
+    case '512':
+      element.classList.add('field-cell--512');
+      break;
+    case '1024':
```
etojeDenys, line 125: you don't need all these cases
```
element.classList.add(`field-cell--${element.innerHTML}`);
```

etojeDenys (changes requested):

src/scripts/main.js
BohdanDymydiuk: ohh, sure

volodymyr-soltys97 (changes requested):
Good job 👍 
Add a favicon on the page 
image

Why does number 2 have the same background color as number 4?
image

The same here, the background color must be are different 
image

volodymyr-soltys97 (approved):
Good job! 👍

PR #947, opened 2024-09-22
add task solution

rberkunova:
- [DEMO LINK](https://rberkunova.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Add a favicon on the page 
image

The start button isn't working now, also fix the messages, you need to hide the message about the lose game 
image

PR #949, opened 2024-09-22
Dev

MaxF1996:
Tests failed, but i can't see a reason (because infinite loop of test). Still, app seems as working. So maybe you can point out the problem.
- [DEMO LINK](https://MaxF1996.github.io/js_2048_game/)

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,29 @@
+name: Test
+
+on:
+  pull_request:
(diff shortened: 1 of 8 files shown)


MaxF1996:
Ok, i find reason. So i will continue work with code.

MaxF1996:
check please, can't see problem

MaxF1996:
image

lerastarynets (approved):
You're on the right track!

PR #955, opened 2024-09-26
add task solution 1.0

StanislavKapytsia:
- [DEMO LINK](https://StanislavKapytsia.github.io/js_2048_game/)

2048 game

Hey! Are you ready for a real hard check of your JavaScript skills, ninja?
If you are still here, let's do it.

In this task, you need to implement the 2048 game like in [this reference](https://play2048.co/)
Don't play for too long! We need you to write the code!

Okay, what do we have?
1) HTML and CSS are already written. You can use it, or implement your own design if you want.
2) Base `Game` class structure is already written too. Extend it with your own methods. Obligatory methods (used in tests):
  - constructor with `initialState` parameter (value is optional, defaults to the empty board)
  - `getState()`
  - `getScore()`
  - `getStatus()`
  - `moveLeft()`
  - `moveRight()`
  - `moveUp()`
  - `moveDown()`
  - `start()`
  - `restart()`

3) Reference.

That's it!

Okay, okay. Also, we have some rules:
1) The game field is 4 x 4
2) Each cell can be empty or contain one of the numbers: 2, 4, 8 ... 2^n
3) The player can move cells with keyboard arrows
4) All the numbers should be moved in the selected direction until all empty cells are filled in
   - 2 equal cells should be merged into a doubled number
   - The merged cell can’t be merged twice during one move
5) The move is possible if at least one cell is changed after the move
6) After move 2 or 4 appears in a random empty cell. 4 probability is 10%
7) When 2048 value is displayed in any cell, win message should be shown.
8) The `game over` message should be shown if there are no more available moves.
9) Hide start message when game starts.
10) Change the `Start` button to `Restart` after the first move.
11) `Restart` button should reset the game to the initial state.
12) Increase score with each move. The score should be increased by the sum of all merged cells.
13) The game consists of 2 main parts:
  - game logic written in `src/modules/Game.class.js` module that exports `Game` class
  - game UI written in `src/index.html` with `main.js` script that need to use `Game` class instance

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ You can change the HTML/CSS layout if you need it.
 ## Deploy and Pull Request
 
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_2048_game/)
+    - [DEMO LINK](https://StanislavKapytsia.github.io/js_2048_game/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
     - Run `npm run test:only -- -n` to run fast test ignoring linter;
--- a/src/index.html
+++ b/src/index.html
@@ -60,11 +60,15 @@
       <div class="message-container">
         <p class="message message-lose hidden">You lose! Restart the game?</p>
         <p class="message message-win hidden">Winner! Congrats! You did it!</p>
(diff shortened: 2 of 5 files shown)


volodymyr-soltys97 (changes requested):
Hi, your demo link isn't working, you need to make a deploy and check it after on the demo link 
image

etojeDenys (approved):
GJ

PR #959, opened 2024-09-27
2048

IrynaZahorodnia:
[DEMO LINK](https://IrynaZahorodnia.github.io/js_2048_game/)


anastasiiavorobiova (changes requested):
Great work!
The new number should be added only if the game field is changed 
Screenshot 2024-09-27 at 17 28 57
Check the ability to continue the game in this case

PR #960, opened 2024-09-27
2048

IrynaZahorodnia:
https://irynazahorodnia.github.io/js_2048_game/


etojeDenys (changes requested):
1. when I press any key(not only the arrows), the cells appear, fix it
2. when you press some arrow key only one new cells should appear
3. you shouldn't add new cells if the board didn't change

https://github.com/user-attachments/assets/a6982409-89c1-4450-829c-16f0a8a5672a

SanyaBratashchuk (approved):

PR #961, opened 2024-09-27
Develop

svitlanak31:
- [DEMO LINK](https://svitlanak31.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,68 +1,185 @@
 'use strict';
 
-/**
- * This class represents the game.
- * Now it has a basic structure, that is needed for testing.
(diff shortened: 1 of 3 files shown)


src/modules/Game.class.js
```diff
@@ -1,68 +1,215 @@
...
+    if (moved) {
+      this.addTile();
+    }
+    this.checkGameStatus();
+  }
+
+  moveRight() {
+    if (this.status !== 'playing') {
+      return;
+    }
```
etojeDenys, line 44: why do you need this check in those functions if you have it here where you call these functions
image

etojeDenys (approved):
good job

PR #963, opened 2024-09-28
solution

oskushnir:
[DEMO LINK](https://oskushnir.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,165 @@
 'use strict';
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
 
-// Write your code here
+const controls = document.querySelector('.controls');
+const score = controls.querySelector('.game-score');
+const startButton = controls.querySelector('.button');
+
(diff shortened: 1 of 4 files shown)


SanyaBratashchuk (changes requested):
image

volodymyr-soltys97 (changes requested):
Hi, there are no styles on the demo link, you need to make a deploy and check it after on the demo 
image

Also, add a favicon on the page 
image

src/scripts/main.js
```diff
@@ -1,7 +1,165 @@
...
+        break;
+
+      case 'ArrowUp':
+        createCellIfGameStarted(game.moveUp(rows));
+
+        score.textContent = game.getScore();
+
+        if (game.getStatus() === 'win') {
+          winMessage.classList.remove('hidden');
+        }
```
etojeDenys, line 120: you can write it only one time after all cases

etojeDenys (changes requested):
only after some extra arrow click the lose message appeared, let's fix that


https://github.com/user-attachments/assets/a4c17e2e-d923-455d-a638-73e8e7c4a185

anastasiiavorobiova (approved):
Great work!

PR #965, opened 2024-10-01
task solution

olshum8:
[DEMO LINK](https://olshum8.github.io/js_2048_game/)

I can't pass the last test, but the actual message about losing is displayed, you can see it on the screenshot below: 
image

I've seen similar errors in the chat, but the solution given by the mentors is to remove the hidden class from the message-lose p, but it's already removed and the message is displayed.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,118 @@
 'use strict';
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+import Game from '../modules/Game.class.js';
 
-// Write your code here
+const game = new Game();
+
+const gameBoard = document.querySelector('.game-field');
+const score = document.querySelector('.game-score');
+const startButton = document.querySelector('#start-button');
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,118 @@
...
+      break;
+  }
+}
+
+// END Update UI
+
+// handle move
+function handleMove(evnt) {
+  switch (evnt.key) {
+    case 'ArrowUp':
```
volodymyr-soltys97, line 72: Create a variables for event keys strings and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Add a favicon on the page 
image

anastasiiavorobiova (changes requested):
Great work! 
New values should be added only if the game field is changed. Consider fixing it

olshum8:
Hi, sorry, I didn't understand what you meant in your last comments. As far as I understand it now, you meant to add a condition that a new cell with a number is not added on the board if the board has not changed after move function. I did it now :)

etojeDenys (approved):
lgtm

PR #969, opened 2024-10-04
game_2048

MartaChobaniuk:
- [DEMO LINK](https://martachobaniuk.github.io/js_2048_game/)


src/modules/Game.class.js
```diff
@@ -1,68 +1,216 @@
...
+        while (newCol.length < 4) {
+          newCol.unshift(0);
+        }
+
+        for (let row = 0; row < 4; row++) {
+          this.state[row][col] = newCol[row];
+        }
+      }
+      this.getRandomCell();
+      this.checkStatus();
```
etojeDenys, line 188: move this logic from all "arrows" function to this function
image
it will simplify the logic

etojeDenys (changes requested):
1. don't create a new cell if nothing has changed on the board

https://github.com/user-attachments/assets/17c6d625-6a46-4c91-8faa-6f7950979ee7

SanyaBratashchuk (approved):

SanyaBratashchuk (approved):

PR #971, opened 2024-10-05
task has been done

u9on:
[DEMO LINK](https://u9on.github.io/js_2048_game/)

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,29 @@
+name: Test
+
+on:
+  pull_request:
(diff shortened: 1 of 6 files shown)


SanyaBratashchuk (changes requested):
test are failed try to fix them

u9on:
> Quote reply
> Reference in new issue



> тест провален попробуйте исправить их

So I don’t understand why they fail, when I run it through VS Code it doesn’t give me anything at all

anastasiiavorobiova (approved):
Great work!

PR #972, opened 2024-10-05
add solution

10hokageee:
https://10hokageee.github.io/js_2048_game/

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,68 +1,266 @@
 'use strict';
 
-/**
- * This class represents the game.
- * Now it has a basic structure, that is needed for testing.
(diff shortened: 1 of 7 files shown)


src/modules/Game.class.js
```diff
@@ -1,68 +1,266 @@
...
+    const moved = this.move('up');
+
+    if (moved) {
+      this.addRandomTile();
+      this.checkState();
+    }
+  }
+
+  moveDown() {
+    if (this.status !== 'playing') {
```
TarasHoliuk, line 108: It makes sense to check in the move function instead of doing it in every function

TarasHoliuk (approved):
Great job!

PR #973, opened 2024-10-07
[Feature] created 2048 game

Arthur-Tolkachov:
[DEMO LINK](https://arthur-tolkachov.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Core.class.js
+++ b/src/modules/Core.class.js
@@ -258,7 +258,12 @@ class Core {
   _shiftLeft() {
     const field = this._getField();
 
-    const { field: newField, score, isWinner } = horizontalShift({ field });
+    const {
+      field: newField,
+      score,
+      isWinner,
+      wasChanged,
+    } = horizontalShift({ field });
 
     this.#updateScore(score);
     this.#updateField(newField);
@@ -266,6 +271,8 @@ class Core {
     if (isWinner) {
       this.#winTheGame();
     }
+
+    return { wasChanged };
   }
 
   _shiftRight() {
@@ -275,6 +282,7 @@ class Core {
       field: newField,
       score,
       isWinner,
+      wasChanged,
     } = horizontalShift({ field, rtl: true });
 
     this.#updateScore(score);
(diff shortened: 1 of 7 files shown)


TarasHoliuk:
If nothing was changed, the new tile shouldn't appear after clicking the arrow button. For example:

Current field:
image

I clicked "left" button (nothing was moved or merged) - and a new tile appeared:

image

src/modules/Core.class.js
```diff
@@ -0,0 +1,319 @@
...
+    field[firstCellPosY][firstCellPosX] = generateCellValue();
+    field[secondCellPosY][secondCellPosX] = generateCellValue();
+
+    this.#updateField(field);
+  }
+
+  #initialRender(rootRef) {
+    rootRef.classList.add('container');
+
+    const headerNode = document.createElement('div');
```
TarasHoliuk, line 72: Your structure is nice, but why just don't you use the initial HTML layout (in index.html) instead of setting it via JS?

src/modules/Game.class.js
```diff
@@ -1,68 +1,144 @@
 'use strict';
+import Core from './Core.class';
+
+import { GAME_STATUSES, KEY_BINDS } from '../lib/constants';
+
+class Game extends Core {
+  #hydrate() {
+    this.controls.startBtn.addEventListener('click', this.start.bind(this));
```
TarasHoliuk, line 8: I like your structure and splitting logic 🔥 👍 

But using HTML elements in the Game class will always lead to tests failing in this task - FYI.

TarasHoliuk (changes requested):
Great job! Fix the issue mentioned in comments

src/modules/Core.class.js
Arthur-Tolkachov: I like this JS-built structure because all you need is just in the box. Just call new Game() class and enjoy, without any additional HTML.

etojeDenys (changes requested):
you should show the message
image
in this case

etojeDenys (approved):

etojeDenys (approved):
🔥

PR #980, opened 2024-10-12
add task solution

mykhailonl:
- [DEMO LINK](https://mykhailonl.github.io/js_2048_game/)


etojeDenys (changes requested):
restart button doesn't work, try to fix that

in this case
image

volodymyr-soltys97 (approved):
Great work! 🔥

PR #981, opened 2024-10-12
add task solution

Syogunok:
- [DEMO LINK](https://Syogunok.github.io/js_2048_game/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. New numbers should appear only when clicking on the arrows
image

2. Add a favicon on the page 
image

etojeDenys (changes requested):
don't create the cell in this case, when the board didn't change
https://github.com/user-attachments/assets/e2219007-40e1-420f-abb4-93ecf93093e9

etojeDenys (approved):
awesome

PR #987, opened 2024-10-17
add solution

an-marryKyslenko:
- [DEMO LINK](https://an-marryKyslenko.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
+    <script
+      src="scripts/main.js"
+      type="module"
+    ></script>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


anastasiiavorobiova (changes requested):
Great work! A user should be able to restart at any point of the game

etojeDenys (changes requested):
looks broken
https://github.com/user-attachments/assets/ea7d5bcd-eb59-4f96-ac26-975b19c9c0e6

etojeDenys (changes requested):
try to press the restart button few times
image
I think it shouldn't be like that

etojeDenys (approved):
good job, but it's better to rewrite the logic with page reloading on the restart click

PR #992, opened 2024-10-21
add task solution

lublubuterbrodi:
https://lublubuterbrodi.github.io/js_2048_game/

image
image


etojeDenys (changes requested):
should be the restart button in this case instead of the start
image

lublubuterbrodi:
Hi, I pushed and deployed changes, in my npm start the button works, but here in the deployment pages it still doesn't. But the code part is done, could you check it please one more time? Thanks)

image

anastasiiavorobiova (approved):

PR #993, opened 2024-10-21
add solution

Tulenty:
[DEMO LINK](https://Tulenty.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,59 @@
 'use strict';
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
+const game = new Game();
 
-// Write your code here
+const startButton = document.querySelector('.start');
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
image
It is better to remove this message if the game has already started

2. in this case, I should have one more move?
image

3. after the lose the board is empty, just restart the game, or show all previous cells
image

anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2024-10-22 at 18 10 18
- Screenshot 2024-10-22 at 18 10 57
- Check styles here

src/scripts/main.js
```diff
@@ -1,7 +1,68 @@
...
+    game.moveRight();
+  }
+
+  if (e.key === 'ArrowUp') {
+    game.moveUp();
+  }
+
+  if (e.key === 'ArrowDown') {
+    game.moveDown();
+  }
```
etojeDenys, line 65: better to rewrite this logic with switch case

etojeDenys (approved):
good job

PR #998, opened 2024-10-22
js_2048_game

viktoriiadereza:
[DEMO LINK](https://viktoriiadereza.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
+    <script
+      src="scripts/main.js"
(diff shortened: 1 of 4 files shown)


anastasiiavorobiova (changes requested):
Great work! A user can`t restart after the loose
Screenshot 2024-10-22 at 22 36 04
When the start is pressed the message should be removed and the button should be changed to the `restart` button

etojeDenys (changes requested):
restart button doesn't work
image

anastasiiavorobiova (approved):
Awesome!

PR #999, opened 2024-10-22
Develop

NazarHaida:
- [DEMO LINK](https://NazarHaida.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,95 @@
 'use strict';
 
 // Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
+const game = new Game();
 
-// Write your code here
+const startBtn = document.querySelector('.button.start');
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,95 @@
...
+      moveMade = true;
+      break;
+    case 'ArrowUp':
+      game.moveUp();
+      moveMade = true;
+      break;
+    case 'ArrowDown':
+      game.moveDown();
+      moveMade = true;
+      break;
```
volodymyr-soltys97, line 33: Create variables for event key words and use it here, also, add the blank line between the cases
```suggestion
    case 'ArrowLeft':
      game.moveLeft();
      moveMade = true;
      break;
      
    case 'ArrowRight':
      game.moveRight();
      moveMade = true;
      break;
      
    case 'ArrowUp':
      game.moveUp();
      moveMade = true;
      break;
      
    case 'ArrowDown':
      game.moveDown();
      moveMade = true;
      break;
```

PR #1000, opened 2024-10-23
beta comit

overnall:
- [DEMO LINK](https://overnall.github.io/js_2048_game/)


anastasiiavorobiova (changes requested):
Great start!
Pressing on the start button doesn't work. Consider fixing it

PR #1007, opened 2024-10-24
Develop

NataliaDavida:
[DEMO LINK](https://NataliaDavida.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -48,6 +48,9 @@ class Game {
 
     if (this.moved) {
       this.addRandomTile();
+      this.getScore();
+      this.updateDOM();
+      this.getStatus();
     }
   }
 
@@ -88,6 +91,9 @@ class Game {
 
     if (this.moved) {
       this.addRandomTile();
+      this.getScore();
+      this.updateDOM();
+      this.getStatus();
     }
   }
 
@@ -130,6 +136,9 @@ class Game {
 
     if (this.moved) {
       this.addRandomTile();
+      this.getScore();
+      this.updateDOM();
+      this.getStatus();
     }
   }
 
@@ -172,6 +181,9 @@ class Game {
 
     if (this.moved) {
       this.addRandomTile();
+      this.getScore();
+      this.updateDOM();
+      this.getStatus();
     }
   }
 
@@ -249,6 +261,12 @@ class Game {
     this.state = this.initialState;
     this.score = 0;
     this.status = 'idle';
+    for (let i = 0; i < this.messages.length; i++) {
+      if (this.messages[i].classList.contains('message-start')) {
+        this.messages[i].classList.remove('hidden');
+      }
+    }
+    
     this.getScore();
     this.updateDOM();
     this.updateButton();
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,55 @@
...
+  function control(e) {
+    if (e.key === 'ArrowLeft') {
+      keyLeft();
+    } else if (e.key === 'ArrowRight') {
+      keyRight();
+    } else if (e.key === 'ArrowUp') {
+      keyUp();
+    } else if (e.key === 'ArrowDown') {
+      keyDown();
+    }
```
etojeDenys, line 23: rewrite it using the switch case

src/scripts/main.js
```diff
@@ -1,7 +1,55 @@
...
+    game.updateDOM();
+    game.getStatus();
+  }
+
+  function keyDown() {
+    game.moveDown();
+    game.getScore();
+    game.updateDOM();
+    game.getStatus();
+  }
```
etojeDenys, line 54:
```
game.getScore();
game.updateDOM();
game.getStatus();
```
you can just move it after the switch case and remove all these functions

etojeDenys (changes requested):
fix your demo link or feel free to ask for some help in the chat how to to that

src/modules/Game.class.js
```diff
@@ -1,68 +1,364 @@
...
+      row.reverse();
+
+      if (row.toString() !== this.state[i].toString()) {
+        this.moved = true;
+      }
+
+      this.state[i] = row;
+    }
+
+    if (this.moved) {
```
TarasHoliuk, line 92: Seems it doesn't work properly

src/modules/Game.class.js
```diff
@@ -1,68 +1,364 @@
...
+      row.reverse();
+
+      for (let j = 0; j < row.length - 1; j++) {
+        if (row[j] === row[j + 1]) {
+          row[j] *= 2;
+          this.score += row[j];
+          row[j + 1] = 0;
+          this.moved = true;
+        }
+      }
```
TarasHoliuk, line 75: This part (almost all method actually) is exactly the same as in moveLeft method

Please, try to reuse some logic between "move" functions

src/modules/Game.class.js
```diff
@@ -1,68 +1,364 @@
...
+          this.messages[i].classList.add('hidden');
+        }
+      }
+
+      this.status = 'playing';
+      this.addRandomTile();
+      this.addRandomTile();
+      this.updateDOM();
+      this.updateButton();
+    } else if (this.status === 'playing') {
```
TarasHoliuk, line 255: What if status is 'lose'?

TarasHoliuk:
1. The new tile shouldn't appear if after clicking the arrow button nothing was changed:

image

Then click 'Left', result - "2" appeared:
image

So you have to check if something changed before adding a new tile

2. The "restart" button doesn't work after loosing game:

image

TarasHoliuk (changes requested):

Anton-Kuchmasov (changes requested):
DEMO link looks broken :(

Image

Would you mind to add

"homepage": "js_2048_game"

to your `package.json` file, save it, add-commit-push and then re-deploy your solution again?

PR #1013, opened 2024-10-31
js game solution

galenwood:
[DEMO LINK](https://galenwood.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,98 @@
 'use strict';
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
+const game = new Game();
 
-// Write your code here
+function refreshTable(state) {
+  const fieldRows = document.querySelectorAll('.field-row');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,98 @@
...
+      break;
+    case 'ArrowRight':
+      e.preventDefault();
+      game.moveRight();
+      break;
+    case 'ArrowDown':
+      e.preventDefault();
+      game.moveDown();
+      break;
+    case 'ArrowLeft':
```
volodymyr-soltys97, line 92: Create a variables for event key words and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Add a favicon on the page 
image

src/scripts/main.js
```diff
@@ -1,7 +1,104 @@
...
+      game.restart();
+  }
+
+  refresh();
+});
+
+document.addEventListener('keydown', (e) => {
+  switch (e.key) {
+    case KEY_UP:
+      e.preventDefault();
```
etojeDenys, line 86: you can write it only one time before the switch/case

etojeDenys (approved):
good job in general

PR #1021, opened 2024-11-07
add task solution

shymdima:
- [DEMO LINK](https://shymdima.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
+    <script
+      type="module"
+      src="scripts/main.js"
+    ></script>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
your demo link doesn't work, please, try to fix and then re-request the review

anastasiiavorobiova (changes requested):
Great work!
- Consider adding proper classes for cells
- Also, the new number shouldn't be added if the game field doesn't change

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Add a favicon on the page 
image

2. Add background color for numbers `2` 
image

3. The first letter must be in upper case, `Restart` 
image

etojeDenys (changes requested):
you don't need to create the cell in this case, add the cell only when you merge something

https://github.com/user-attachments/assets/736f79cc-b632-40d8-beb2-27ed5fc02a7a

VitaliyBondarenko1982 (approved):
Looks good

PR #1030, opened 2024-11-11
Develop

IrynaMariiko00:
- [DEMO LINK](https://IrynaMariiko00.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -7,10 +7,11 @@
  */
 const IDLE = 'idle';
 const PLAYING = 'playing';
-const WIN = 'win';
-const LOSE = 'lose';
 
-class Game {
+export const WIN = 'win';
+export const LOSE = 'lose';
+
+export class Game {
   /**
    * Creates a new game instance.
    *
@@ -25,7 +26,7 @@ class Game {
    * If passed, the board will be initialized with the provided
    * initial state.
    */
-  constructor(initialState) {
+  constructor(initialState, winCallback, loseCallback) {
     this.board = initialState || [
       [0, 0, 0, 0],
       [0, 0, 0, 0],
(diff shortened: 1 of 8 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,95 @@
...
+  startButton.textContent = 'Restart';
+  startButton.classList.remove('start');
+  startButton.classList.add('restart');
+
+  cells.forEach((cell) => {
+    cell.classList.remove('hidden');
+  });
+
+  if (game.status === WIN || game.status === LOSE) {
+    hideMessage();
```
TarasHoliuk, line 72: You should remove start message:

image

TarasHoliuk: Also, add margin to Lose message
image

src/modules/Game.class.js
```diff
@@ -20,26 +26,358 @@ class Game {
...
+
+    // Затримка для застосування анімації збільшення
+    setTimeout(() => {
+      // Додаємо клас 'merged' для анімації збільшення
+      toCell.classList.add('merged');
+
+      // Затримка для видалення класу 'merged' після завершення анімації
+      setTimeout(() => {
+        toCell.classList.remove('merged');
+      }, 300); // Час анімації (300ms)
```
TarasHoliuk, line 69: If you change some property in `merged` class you should add transition for this property in scss file
Now it doesn't work (but not required in the task - so it's only a suggestion)

src/modules/Game.class.js
```diff
@@ -20,26 +26,358 @@ class Game {
...
+        const value = this.board[row][col];
+        const previousValue = cell.textContent;
+
+        cell.classList.remove(`field-cell--${previousValue}`);
+
+        if (value === 0) {
+          cell.textContent = '';
+        } else {
+          cell.classList.add(`field-cell--${value}`);
+          cell.textContent = value; // оновлюємо текст
```
TarasHoliuk, line 91: it's clear without comment 🙂 And it's related to 90% of comments - no need to comment smth if it's clear from the code. If there is complicated logic, unusual behavior, etc. we need to add comments

TarasHoliuk (changes requested):
GJ 👍 Let's improve it a bit)

src/modules/Game.class.js
IrynaMariiko00: oh, sorry, I wrote it for myself and forgot to delete it

src/scripts/main.js
```diff
@@ -1,7 +1,95 @@
...
+function saveScore(sscore) {
+  const currentRecord = localStorage.getItem('highScore');
+
+  if (!currentRecord || sscore > currentRecord) {
+    localStorage.setItem('highScore', sscore);
+    displayRecord(); // Оновлюємо рекорд на екрані
+  }
+}
+
+// Функція для відображення рекорду
```
volodymyr-soltys97, line 43: Remove all comments
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
There is no commit from the previous review, you need to commit and push your changes
The text in this button should be centered vertically
image

When we restart the game you need to reset score 
image

src/styles/main.scss
```diff
@@ -177,11 +183,56 @@ h1 {
...
 .message-win {
-  background: #edc22e;
+  background: #2669b5;
   color: #f9f6f2;
 }
 
+.message-lose {
+  width: 100%;
+  box-sizing: border-box;
+  height: 60px !important;
```
volodymyr-soltys97, line 193: Don't use important, it's a bad practice
```suggestion
  height: 60px;
```

volodymyr-soltys97 (changes requested):
Almost done!
This comment still not fixed from the previous review 
image

volodymyr-soltys97 (changes requested):
.

etojeDenys (approved):
good job in general!

IrynaMariiko00:
I have improved my code again. Could you look at it, please?

PR #1045, opened 2024-11-25
add task solution

AndriiZhovtko:
- [DEMO LINK](https://AndriiZhovtko.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,68 +1,266 @@
-'use strict';
-
-/**
- * This class represents the game.
- * Now it has a basic structure, that is needed for testing.
- * Feel free to add more props and methods if needed.
- */
 class Game {
-  /**
-   * Creates a new game instance.
-   *
-   * @param {number[][]} initialState
-   * The initial state of the board.
-   * @default
-   * [[0, 0, 0, 0],
-   *  [0, 0, 0, 0],
-   *  [0, 0, 0, 0],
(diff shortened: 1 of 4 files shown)


src/modules/Game.class.js
```diff
@@ -1,68 +1,266 @@
...
+  start() {
+    if (this.status === 'idle') {
+      this.status = 'playing';
+      this.addRandomTile();
+      this.addRandomTile();
+    }
+  }
+
+  restart() {
+    if (this.status === 'playing') {
```
TarasHoliuk, line 246: What if status is `'lose'`?

src/modules/Game.class.js
```diff
@@ -1,68 +1,266 @@
...
+    if (moved) {
+      this.addRandomTile();
+      this.checkGameState();
+    }
+  }
+
+  move(direction) {
+    const originalState = this.copyState(this.state);
+
+    const combineRow = (row) => {
```
TarasHoliuk, line 95: you can define it as a separate method instead of creating a new function inside `move` method

and similar for functions below

TarasHoliuk (changes requested):
Great job!

But the restart button doesn't work if a user lose the game

SanyaBratashchuk (changes requested):
restart button still not working 
image

etojeDenys (approved):
lgtm

PR #1048, opened 2024-11-26
solution

sophiasph:
[DEMO LINK](https://sophiasph.github.io/js_2048_game/)


volodymyr-soltys97 (changes requested):
Almost done!
Hide this message after starting the game 
image

volodymyr-soltys97 (changes requested):
Hi, the demo link still not fixed
You don't need to completely delete this message, you only need to hide it after we start the game.
image

PR #1053, opened 2024-11-28
2048

Dushkaaa-BabyBoss:


src/modules/Game.class.js
```diff
@@ -23,22 +23,156 @@ class Game {
...
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+    ];
+
+    this.board = initialState || JSON.parse(JSON.stringify(this.defaultState));
+    this.score = 0; // Початкови рахунок
+    this.status = 'idle'; // Початковий статус гри
+  }
+
+  moveLeft() {
```
TarasHoliuk, line 40: it works even before I click "start" button - it should be fixed:

image

src/modules/Game.class.js
```diff
@@ -23,22 +23,156 @@ class Game {
...
+      newRow = newRow.filter((value) => value !== 0);
+
+      while (newRow.length < 4) {
+        newRow.push(0);
+      }
+
+      this.board[i] = newRow;
+    }
+
+    this.addRandomTile();
```
TarasHoliuk, line 63: You should not add a new tile if nothing changed:

image

Then I clicked left, as result - the new tile appeared:
image

src/modules/Game.class.js
```diff
@@ -50,17 +184,81 @@ class Game {
...
+
+      this.board[row][col] = Math.random() < 0.9 ? 2 : 4;
+    }
+  }
 
   /**
    * Resets the game.
    */
-  restart() {}
+  restart() {
```
TarasHoliuk, line 259: after clicking restart, the button text is "Start", but new tiles appeared and I can play

src/modules/Game.class.js
```diff
@@ -23,22 +23,156 @@ class Game {
...
+      [0, 0, 0, 0],
+    ];
+
+    this.board = initialState || JSON.parse(JSON.stringify(this.defaultState));
+    this.score = 0; // Початкови рахунок
+    this.status = 'idle'; // Початковий статус гри
+  }
+
+  moveLeft() {
+    this.checkLose();
```
TarasHoliuk, line 41: it should be done in after move - same as checkWin (no need to do it in each "move" method)

src/modules/Game.class.js
```diff
@@ -23,22 +23,156 @@ class Game {
...
+  moveRight() {
+    this.checkLose();
+
+    for (let i = 0; i < 4; i++) {
+      let newRow = this.board[i]
+        .slice()
+        .reverse()
+        .filter((value) => value !== 0);
+
+      for (let j = 0; j < newRow.length - 1; j++) {
```
TarasHoliuk, line 74: code in this method is almost the same, but you do reverse before and after manipulations. Try to add reusable code to a separate function and reuse it instead of duplication

TarasHoliuk (changes requested):
Try to fix tests

src/modules/Game.class.js
```diff
@@ -23,22 +23,201 @@ class Game {
   constructor(initialState) {
     // eslint-disable-next-line no-console
     console.log(initialState);
+
+    // Ініціалізуємо гральну дошку
```
volodymyr-soltys97, line 27: Remove all comments
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job!
You need to add favicon on the page 
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
When game not started, shouldn't any tiles on board.
Screenshot 2024-12-02 at 17 31 42

TarasHoliuk (approved):
LGTM, great job!

PR #1057, opened 2024-12-03
2048

KarinaKoliada:
- [DEMO LINK](https://KarinaKoliada.github.io/js_2048_game/)


src/modules/Game.class.js
```diff
@@ -1,68 +1,403 @@
...
+            this.state[row][tile + moveTile + 1] === this.state[row][tile]
+          ) {
+            moveTile++;
+            merge = true;
+          }
+
+          if (moveTile > 0) {
+            if (!didJustMerged && merge) {
+              didJustMerged = true;
+
```
volodymyr-soltys97, line 146: This logic is quite similar and it is duplicated in every method, try to move it to a helper function and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #1060, opened 2024-12-04
Solution

AndreaTkachuk:


AndreaTkachuk:
- [DEMO LINK](https://AndreaTkachuk.github.io/js_2048_game/)

etojeDenys (changes requested):
you don't need to create new cell if the board didn't change

https://github.com/user-attachments/assets/10429e6f-d7fe-479f-8199-4cfbf2387059


2. I still should have opportunity to move
image

AndreaTkachuk:
- [DEMO LINK](https://AndreaTkachuk.github.io/js_2048_game/)

etojeDenys (changes requested):
the lose message should disappear after the restart
image

2. you need to rewrite your logic how to detect if the user lose the game
image
in this case only after moveDown user will see the lose message and it's incorrect

VitaliyBondarenko1982 (approved):
Looks good

PR #1065, opened 2024-12-08
Develop

shelyaa:
https://shelyaa.github.io/js_2048_game/


etojeDenys (changes requested):
you don't need to add a new cell if the board didn't change

https://github.com/user-attachments/assets/f46c3c67-5e6f-4c70-9648-4ee9a2b0d804

etojeDenys (approved):

PR #1072, opened 2024-12-11
resolution game_2048

14lucas-mendes:
[DEMO LINK](https://14lucas-mendes.github.io/js_2048_game/)


pedro-ruas (changes requested):
Issues with this code:
- Looks AI made
- Is failing all tests (should pass all)
- Game does not work (no numbers shown)
- Restart does not work

src/index.html
```diff
@@ -19,13 +19,26 @@ <h1>2048</h1>
...
+            <span
+              class="game-score"
+              id="score"
+            >
+              0
+            </span>
           </p>
-          <button class="button start">Start</button>
+          <button
+            class="button start"
```
pedro-ruas, line 30: I believe there should also be a button-restart, right? It doesn't work when the game is lost

src/modules/Game.class.js
```diff
@@ -6,63 +6,416 @@
...
+
+  init() {
+    // Configura os eventos
+    this.bindEvents();
+    // Atualiza o display inicial
+    this.updateDisplay();
+  }
+
+  // Metodo para reiniciar o jogo
+  restart() {
```
pedro-ruas, line 49: Are you using this function?

src/modules/Game.class.js
```diff
@@ -6,63 +6,416 @@
...
+    }
+  }
+
+  bindEvents() {
+    // Evento do botão start
+    const startButton = document.getElementById(this.startId);
+
+    if (startButton) {
+      startButton.addEventListener('click', () => {
+        this.start();
```
pedro-ruas: Maybe you could pass the restart function here too

pedro-ruas (changes requested):
Restart is not working, because there is no restart button
 Game needs to be initialized twice to run

pedro-ruas (approved):

PR #1074, opened 2024-12-11
Solution

DianaKadimbekova:
- [DEMO LINK](https://DianaKadimbekova.github.io/js_2048_game/)

instead
const transposed = [];

    for (let col = 0; col < this.board[0].length; col++) {
      const newRow = this.board.map((row) => row[col]);

      transposed.push(newRow);
    }

i had
const transposed = this.board[0].map((_, i) =>
      this.board.map((row) => row[i]),
    );
but the terminal wouldn't let me through.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ You can change the HTML/CSS layout if you need it.
 ## Deploy and Pull Request
 
 1. Replace `<your_account>` with your Github username in the link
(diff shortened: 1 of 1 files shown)


etojeDenys (changes requested):
feel free to ask for some help in the chat with that terminal

etojeDenys (approved):
🔥

PR #1075, opened 2024-12-12
add solution

clavigo:
[DEMO LINK](https://clavigo.github.io/js_2048_game/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,51 +7,56 @@ In this task, you need to implement the 2048 game like in [this reference](https
 Don't play for too long! We need you to write the code!
 
 Okay, what do we have?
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
you don't need to create a new cell if the board didn't change

https://github.com/user-attachments/assets/bfffbb67-19ab-466c-8ffe-e27a40a3bf76

etojeDenys (changes requested):
I can still reproduce that error

clavigo:
> I can still reproduce that issue

I forgot to update the deploy...

etojeDenys (approved):
🔥

PR #1093, opened 2024-12-22
Game 2048

pdabizha:
[DEMO LINK](https://pdabizha.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -20,25 +20,113 @@ class Game {
    * If passed, the board will be initialized with the provided
    * initial state.
    */
-  constructor(initialState) {
-    // eslint-disable-next-line no-console
-    console.log(initialState);
+  constructor(
+    initialState = [
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+    ],
+  ) {
+    this.initialState = initialState;
+    this.board = initialState.map((row) => [...row]);
+    this.score = 0;
+    this.isGameStarted = false;
   }
 
-  moveLeft() {}
-  moveRight() {}
-  moveUp() {}
-  moveDown() {}
+  moveLeft() {
+    const previousBoard = this.getState();
+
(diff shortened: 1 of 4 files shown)


src/modules/Game.class.js
```diff
@@ -50,19 +138,139 @@ class Game {
...
+    if (!this.isGameStarted) {
+      return 'idle';
+    }
+
+    const hasWinningCell = this.board.some((row) => {
+      return row.some((cell) => cell === 2048);
+    });
+
+    if (hasWinningCell) {
+      return 'win'; // Игрок выиграл
```
volodymyr-soltys97, line 151: Remove comments
```suggestion
      return 'win';
```

src/modules/Game.class.js
```diff
@@ -50,19 +138,139 @@ class Game {
...
+
+  joinValues(arr, targetLength) {
+    const result = [];
+    let scoreIncrease = 0;
+
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i] === arr[i + 1]) {
+        const mergedValue = arr[i] * 2;
+
+        result.push(mergedValue); // Удваиваем значение
```
volodymyr-soltys97, line 241: Remove comments everywhere 
```suggestion
        result.push(mergedValue);
```

src/modules/Game.class.js
```diff
@@ -20,25 +20,113 @@ class Game {
...
+      transposedBoard[index] = combinedRow;
+    });
+
+    this.board = this.transpose(transposedBoard);
+
+    if (this.isBoardChanged(previousBoard, this.board)) {
+      const randomCell = this.getRandomCell(1);
+
+      this.addNewCell(randomCell);
+    }
```
volodymyr-soltys97, line 91: This logic is similar in every method, so it would be better to move it to a helper function and use it everywhere.

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Add favicon on the page 
image

etojeDenys (approved):
amazing

PR #1094, opened 2024-12-25
Develop

taniakolesnik:
[DEMO LINK](https://taniakolesnik.github.io/js_2048_game/)
Please note that some test fails as added random new cells cannot make test pass.

- should have a state with two cells with values 2 or 4 - UI all good. Can you please advise?
- should generate 4 less often than 2 and should generate the board with cells in random positions - can you please advise about these ones?
- if the move does not change the board - my game does not have state like this.

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,29 @@
+name: Test
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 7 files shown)


etojeDenys (changes requested):
restart button works strange

https://github.com/user-attachments/assets/20477e1e-c299-4651-9c3b-5c39d317dd6d


2. you don't need to add a new cell if the board didn't change

https://github.com/user-attachments/assets/2132fd46-694f-4514-9274-56802f54eade

PR #1102, opened 2024-12-31
test_1

igols:


vadiimvooo (changes requested):
Good job, but your game is not working properly at the moment. For example:
I had this position
image
And in fact when I press down button no actions should be done, but instead I got
image

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Add a favicon on the page 
image

2. You need to hide this message after started the game 
image

etojeDenys (changes requested):
don't forget to redeploy your demo link, because I still can reproduce previous comment

VitaliyBondarenko1982 (approved):
Looks good.

PR #1123, opened 2025-01-13
add solution

LX-y-ka:
- [DEMO LINK](https://LX-y-ka.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
+    <script
+      src="scripts/main.js"
+      type="module"
+    ></script>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


vadiimvooo (changes requested):
Seems like you forgot to deploy your solution. Please feel free to ask for help in fe_chat.

etojeDenys (changes requested):
1. do not add new cell if the board didn't change

https://github.com/user-attachments/assets/9181d4c9-a03c-4dd3-ae4b-9a9f21f7f8d1


2. add two cells on start
image

3. I don't lose, still should have the opportunity to moveUp
image

vadiimvooo (changes requested):
Good job, there are a few comments to fix:
1. when game starts add 2 cells instead of one
2. At this position bottom click should not create new cell (as other cells haven't merged)
image
3. We still have 1 possible move but game shows that we lost, please fix it.
image

vadiimvooo (approved):
Well done

PR #1137, opened 2025-01-21
Develop

iirk1:
https://iirk1.github.io/js_2048_game/


vadiimvooo (changes requested):
Good job, but your game is not working properly.
At this position down click should do nothing
image
But instead it add new cell
image
Also message should dissapear when game was started.
image

anastasiiavorobiova (approved):

PR #1142, opened 2025-01-22
solution

KateBerd:
[DEMO LINK](https://KateBerd.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -6,6 +6,12 @@
  * Now it has a basic structure, that is needed for testing.
  * Feel free to add more props and methods if needed.
  */
+const {
+  moveRowLeft,
+  moveRowRight,
+  transpose,
+  areBoardsEqual,
+} = require('../helpers/moveCells.js');
 
 class Game {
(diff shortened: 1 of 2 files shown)


vadiimvooo (changes requested):
Good job, but there is one comment to fix. Please feel free to ask for help in fe_chat.
On each move you should create only one tile, but now you sometimes generating 2.
image

src/modules/Game.class.js
```diff
@@ -6,63 +7,216 @@
...
+      }
+    }
+
+    const mergedRow = filteredRow.filter((cell) => cell !== 0);
+
+    while (mergedRow.length < 4) {
+      mergedRow.push(0);
+    }
+
+    return mergedRow;
```
volodymyr-soltys97, line 95: This logic is similar in every method, try to put it in a helper function and use it everywhere.

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code
Add a favicon on the page 
image

vadiimvooo (approved):
LGTM

PR #1153, opened 2025-01-26
help

VadymBryvka:
[DEMO](https://vadymbryvka.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,6 @@
         </p>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


volodymyr-soltys97 (changes requested):
Hi, there is no demo link, also, you need to add demo link to the PR description 
image

Also, you need to create a new branch for your solution in future 
image

etojeDenys (changes requested):
fix your demo link and then re-request the review
image

PR #1155, opened 2025-01-28
add solution

Larysa1387:
- [DEMO LINK](https://Larysa1387.github.io/js_2048_game/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,51 +7,56 @@ In this task, you need to implement the 2048 game like in [this reference](https
 Don't play for too long! We need you to write the code!
 
 Okay, what do we have?
-1) HTML and CSS are already written. You can use it, or implement your own design if you want.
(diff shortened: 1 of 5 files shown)


vadiimvooo (changes requested):
Good job, there are a few changes to fix:
1. Align of cells is not the same for empty and filled values.
image
2. You should not add new cells if there is no merge available:
For example in this position on move down there should not be added new cell.
image

Larysa1387:
Sorry, I didn`t catch this : Align of cells is not the same for empty and filled values....
It was linter issue in VScode in scss with .field-cell { vertical-align: center;}.. And i fixed it, like {vertical-align: middle;}. But it only has effect on aligning inside cell. Other styles weren't  changed

etojeDenys (approved):

PR #1163, opened 2025-02-01
add task solution

amir-al-mohamad:
[DEMO LINK](https://amir-al-mohamad.github.io/js_2048_game/)


etojeDenys (changes requested):
you don't need to create a cell if the board didn't change

https://github.com/user-attachments/assets/2f78fd4f-8c6d-48b6-83ed-b83886a2fae6

etojeDenys (approved):
good job

PR #1179, opened 2025-02-06
js_2048

AlbinaAlbi:


vadiimvooo (changes requested):
Good job, but seems like your start button is not working, please fix it. Don't hesitate to ask for help in fe_chat.

etojeDenys (changes requested):
restart button should appear after the start click
image


do not create a new cell if the board didn't change

https://github.com/user-attachments/assets/75ee6158-a2fb-4b30-891a-3f28af5ffd47

vadiimvooo (changes requested):
Good job, there are a few fixed to do:
1. Restart button should appear instead of start but not as separate element.
image
2, At this position on bottom click nothing should change, but instead you adding new cell, please fix it.
image

vadiimvooo (changes requested):
Good job, but there is one problem remains. Please feel free to ask for help in fe_chat
No cells should be created if the move wasn't made. However, in this case, a cell is still created when the button is clicked.
image

PR #1193, opened 2025-02-12
solution

CNegruzzi:
[DEMO LINK](https://CNegruzzi.github.io/js_2048_game/)


vadiimvooo (changes requested):
Good job, but there is an issue which needs to be fixed:
1. When cells wasn't merged by move, you should not add new one on table.
image
But instead you do
image

etojeDenys (changes requested):
after the loss, if I click the restart button, it always shows the start button
image


2. do not add the cell if the board didn't change

https://github.com/user-attachments/assets/e8fccadd-6d08-4e75-ae0c-bba65aea9a4c

etojeDenys (changes requested):
it's still not fixed

https://github.com/user-attachments/assets/3fee89a6-e1a5-47c8-9d10-d41185c659a2

volodymyr-soltys97 (approved):
Good job 👍

PR #1203, opened 2025-02-18
add solution

ZorosMG:
[DEMO LINK](https://ZorosMG.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -84,6 +84,7 @@ class Game {
 
   moveLeft() {
     let moved = false;
+    const oldBoard = JSON.stringify(this.board);
 
     this.board.forEach((row, rowIndex) => {
       const { newRow, moved: rowMoved } = this.slideAndMergeRow(row);
@@ -95,7 +96,9 @@ class Game {
     });
 
     if (moved) {
-      this.addRandomTile();
+      if (JSON.stringify(this.board) !== oldBoard) {
+        this.addRandomTile();
+      }
     }
 
     this.checkGameStatus();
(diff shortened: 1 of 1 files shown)


src/modules/Game.class.js
```diff
@@ -1,68 +1,265 @@
...
+    // If there are any empty cells, place a random tile
+    if (emptyCells.length > 0) {
+      const [x, y] = emptyCells[Math.floor(Math.random() * emptyCells.length)];
+      const newValue = Math.random() < 0.9 ? 2 : 4;
+
+      this.board[x][y] = newValue;
+    }
+  }
+
+  // Check if a move is valid (if any tile moves or merges)
```
volodymyr-soltys97, line 71: Remove comments

src/modules/Game.class.js
```diff
@@ -1,68 +1,265 @@
...
+      }
+      this.board[rowIndex] = newRow;
+    });
+
+    if (moved) {
+      this.addRandomTile();
+    }
+
+    this.checkGameStatus();
+    this.updateUI();
```
volodymyr-soltys97, line 107: This logic is similar in every method, try to put it in a helper function and use it everywhere.

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
Add a favicon on the page 
image

anastasiiavorobiova (changes requested):
Great work!
If the game field doesn`t change the new number shouldn't be added
Also, the Game shouldn't be over if there are available options to make the next step

PR #1209, opened 2025-02-19
Firs 2048

Oksigenka:
- [DEMO LINK](https://Oksigenka.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -21,24 +21,197 @@ class Game {
    * initial state.
    */
   constructor(initialState) {
-    // eslint-disable-next-line no-console
-    console.log(initialState);
+    this.score = 0;
+    this.scoreVal = document.querySelector('.game-score');
+
+    this.field = [
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+    ];
+    this.isStart = true;
   }
 
-  moveLeft() {}
-  moveRight() {}
-  moveUp() {}
-  moveDown() {}
+  shiftLeft() {
+    let empty = -1;
+
+    for (let i = 0; i < 4; i++) {
+      empty = -1;
+
+      for (let j = 0; j < 4; j++) {
+        if (this.field[i][j] === 0 && empty === -1) {
+          empty = j;
+        } else if (this.field[i][j] !== 0 && empty !== -1) {
+          this.field[i][empty] = this.field[i][j];
+          this.field[i][j] = 0;
+          empty++;
+        }
+      }
+    }
+  }
+
+  moveLeft() {
+    this.shiftLeft();
+
+    for (let i = 0; i < 4; i++) {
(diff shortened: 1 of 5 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

2. The game board should be an empty by default 
image

TarasHoliuk:
1. The new tile is added when I click on any button - should be fixed

2. Do not add a new tile when nothing has changed (no one tile was moved or merged. I.e. - state remain unchanged) after clicking arrow button

src/modules/Game.class.js
```diff
@@ -50,19 +223,135 @@ class Game {
...
-  getStatus() {}
+
+  getStatus() {
+    let emptyField = true;
+
+    for (let i = 0; i < 4; i++) {
+      for (let j = 0; j < 4; j++) {
+        if (this.field[i][j] === 2048) {
+          document.querySelector('.message-win').style.display = 'block';
+          document.removeEventListener('keydown', this.keyMove);
```
TarasHoliuk, line 234: You shouldn't do DOM manipulation inside of the Game class. Do all such things in main.js

src/modules/Game.class.js
```diff
@@ -50,19 +223,135 @@ class Game {
...
+      const { x, y } =
+        emptyCells[Math.floor(Math.random() * emptyCells.length)];
+
+      if (this.field[x][y] === 0) {
+        this.field[x][y] = Math.random() < 0.9 ? 2 : 4;
+
+        if (this.isStart === true) {
+          const z = Math.floor(Math.random() * 3) + 1;
+
+          if (x + z <= 3 && y + z <= 3) {
```
TarasHoliuk, line 303: create a function to add a cell in a random empty position and use it everywhere you need - call it twice when the game starts or call it once ofter player's move

src/modules/Game.class.js
```diff
@@ -50,19 +223,135 @@ class Game {
...
+            this.field[x + z][y + z] = Math.random() < 0.9 ? 2 : 4;
+          }
+
+          this.isStart = false;
+        }
+      }
+    } else {
+      this.getStatus();
```
TarasHoliuk, line 311: this function has the wrong name or does the wrong thing. Get status method should only return status

src/modules/Game.class.js
```diff
@@ -50,19 +223,135 @@ class Game {
...
+    }
+  }
+
+  gameOver() {
```
TarasHoliuk, line 315: the wrong name - it checks whether there is available move

TarasHoliuk (changes requested):
Try to not duplicate your code and divide it into a separate reusable functions with one thing they really do (when name correctly reflects what the function does)

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. After clicking on start game only 2 numbers should appear on the board
image

2. The same with restart logic 
image

PR #1231, opened 2025-03-01
add task solution

kovaden414:
- [DEMO LINK](https://kovaden414.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,68 +1,187 @@
 'use strict';
 
-/**
- * This class represents the game.
- * Now it has a basic structure, that is needed for testing.
- * Feel free to add more props and methods if needed.
- */
 class Game {
-  /**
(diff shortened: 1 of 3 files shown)


src/modules/Game.class.js
```diff
@@ -1,68 +1,187 @@
...
+        this.board[row][col] = column[row];
+      }
+    }
+
+    if (oldBoard !== JSON.stringify(this.board)) {
+      this.addRandomTile();
+    }
+
+    this.endGame();
+  }
```
volodymyr-soltys97, line 64: This logic is similar in every method, try to put it in a helper function and use it everywhere.

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

vadiimvooo (approved):
Well done

PR #1236, opened 2025-03-04
solution

denisbielykh:
https://denisbielykh.github.io/js_2048_game/


volodymyr-soltys97 (changes requested):
Hi, you need to fix the demo link. Also add a favicon to the page 
image

PR #1267, opened 2025-03-23
add task solution

hetskoweb:
[DEMO LINK](https://hetskoweb.github.io/js_2048_game/)


etojeDenys (changes requested):
you don't need to create a cell if the board didn't change after move

https://github.com/user-attachments/assets/dfa19f77-c6f9-4ac2-8edc-dca97ea815a5

src/modules/Game.class.js
```diff
@@ -20,25 +20,249 @@ class Game {
...
+      this.getState();
+      this.checkFirstMove();
+      this.getScore();
+
+      const gameStatus = this.getStatus();
+      const messageWin = document.querySelector('.message-win');
+      const messageLose = document.querySelector('.message-lose');
+
+      if (gameStatus === 'win') {
+        messageWin.classList.remove('hidden');
```
volodymyr-soltys97, line 87: This logic is similar in every method, you need to save it in a helper function and use it everywhere.

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
The comment from the previous review still not fixed

Add a favicon to the page 
image

vadiimvooo (approved):
Well done

PR #1269, opened 2025-03-27
2048 game

annaabramovaa:
- [DEMO LINK](https://annaabramovaa.github.io/js_2048_game/)


vadiimvooo (changes requested):
Good job, there are a few comments to fix. Please feel free to ask for help in fe_chat.
1. You should add new cells if no move was done. (For example here you don't need to add new cell on bottom move however you do..)
image
2. You should show lose message if there are no available moves
image

etojeDenys (approved):

PR #1284, opened 2025-04-04
solution-2048

romkaqb:
[DEMO LINK](https://romkaqb.github.io/js_2048_game/)


vadiimvooo (changes requested):
Good job, there are few changes to do. Please feel free to ask for help in fe_chat.
1. Show empty cells when game is not started.
image
2. When no cell merge or no move were done you shouldn't create any cell. (For example below if bottom click was done, there should not appear any new cell)
image
But it appears
image
3. Add favicon to your page

vadiimvooo (changes requested):
Almost done. Just on bug at the moment. You don't show lose message after no available moves are exists 
image
instead it's required to click once again and after it will appear.
image

etojeDenys (approved):

PR #1300, opened 2025-04-14
Done!

gmaikovych:
- [DEMO LINK](https://gmaikovych.github.io/js_2048_game/)


etojeDenys (changes requested):
1. I didn't lose, still should have the opportunity to move up or down
image

2. Add a new cell when the board changes, instead of when any button is clicked.

anastasiiavorobiova (changes requested):
Great work!
It shows an additional number after the start
Screenshot 2025-04-15 at 15 28 26

gmaikovych:
Thanks! That additional number was printout in initial state. Fixed.

anastasiiavorobiova (changes requested):
The new number should be added only if the game field is changed

gmaikovych:
Done,
still I'm not sure if understood correctly when exactly new value should be added. 
So if it's wrong again, please, explain the logic.

etojeDenys (approved):

PR #1316, opened 2025-04-24
Add js_2028_game implementation

YaroslavYarmoshyk:


etojeDenys (changes requested):
https://github.com/user-attachments/assets/16f4e2c5-97a8-48b1-af78-d016874984de

do not add the new cells if the board is not changed

etojeDenys (approved):

PR #1331, opened 2025-05-01
add task solution

IvanIlnytskyy:
- [DEMO LINK](https://IvanIlnytskyy.github.io/js_2048_game/)


etojeDenys (changes requested):
your demo link doesn't work, try to fix that and then re-request the review

src/modules/Game.class.js
```diff
@@ -20,25 +20,224 @@ class Game {
...
+          if (this.board[i][this.size - 1 - j] !== row[j]) {
+            moved = true;
+          }
+        }
+      }
+
+      while (newRow.length < this.size) {
+        newRow.push(0);
+      }
+
```
volodymyr-soltys97: This logic is similar in every method, try to put it in a helper function and use it everywhere.

src/modules/Game.class.js
```diff
@@ -20,25 +20,224 @@ class Game {
...
   }
+  moveUp() {
+    if (this.status !== 'playing') {
+      return;
+    }
+
+    let moved = false;
+
+    for (let col = 0; col < this.size; col++) {
+      // ⬆️ Зчитуємо колонку
```
volodymyr-soltys97: These comments looks from ChatGPT, it's a bad practice, you need to write your own solution and remove all comments

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

2. Button on the right hasn't indentation
image

volodymyr-soltys97 (approved):
GJ!

PR #1338, opened 2025-05-04
Game_2048

aCTORr8:
- [DEMO LINK](https://aCTORr8.github.io/js_2048_game/)


src/modules/Game.class.js
```diff
@@ -20,25 +20,229 @@ class Game {
...
+          break;
+        }
+      }
+
+      this.board[r] = newRow;
+    }
+
+    if (moved) {
+      this.afterMove();
+    }
```
volodymyr-soltys97: This logic is similar in every method, try to put it in a helper function and use it everywhere.

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

etojeDenys (approved):
good job overall

PR #1363, opened 2025-05-18
add task solution

lerika7:
https://lerika7.github.io/js_2048_game/


src/modules/Game.class.js
```diff
@@ -20,25 +20,233 @@ class Game {
...
+
+    if (moved) {
+      this.state = newState;
+      this.addRandomTile();
+      this.score += totalScore;
+      this.checkWin();
+      this.checkLose();
+    }
+
+    return moved;
```
volodymyr-soltys97: You have similar logic in every method and it is duplicated, try moving it to a helper function and using it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #1371, opened 2025-05-26
2048 solution

PavloStelmaschuk:
- [DEMO LINK](https://PavloStelmaschuk.github.io/js_2048_game/)


src/modules/Game.class.js
```diff
@@ -1,68 +1,234 @@
...
+    const [row, col] =
+      emptyCells[Math.floor(Math.random() * emptyCells.length)];
+
+    this.board[row][col] = Math.random() < 0.9 ? 2 : 4;
+  }
+
+  checkGameStatus() {
+    for (let i = 0; i < this.size; i++) {
+      for (let j = 0; j < this.size; j++) {
+        if (this.board[i][j] === 2056) {
```
TarasHoliuk: 2048 is winning number

TarasHoliuk (changes requested):
Great job! Let's improve one thing: you shouldn't add a new tile if nothing changed after the arrow button click

TarasHoliuk (changes requested):
.

etojeDenys (approved):
good job. however, it would be better to use a switch case here
image

PR #1379, opened 2025-05-28
Solution

Uandrik:
- [DEMO LINK](https://Uandrik.github.io/js_2048_game/)


maxim2310 (changes requested):

1. remove this block when game is started 
image
2. if i can't move down and click nothing should happen, but now you create new cell 
image
image
3. event move left doesn't work 
4. you ask user if he wants to restart game but game is not restarted, you need to click restart again, i would remove this modal at all and just restart game 
image

PR #1389, opened 2025-06-01
add solution

X-Rayon-X:
- [DEMO LINK](https://X-Rayon-X.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
+    <script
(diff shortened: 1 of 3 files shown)


VitaliyBondarenko1982 (changes requested):
Good job.

After click on start button should appear two new cells.
Please, compare with reference game.

https://github.com/user-attachments/assets/aacb62c7-e3c0-4359-bde6-eb062fb9c47f

etojeDenys (approved):
Good job! However, it would be better to use a switch case
image

PR #1397, opened 2025-06-05
add task solution

combine143:
https://combine143.github.io/js_2048_game/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
+    <script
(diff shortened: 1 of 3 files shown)


anastasiiavorobiova (changes requested):
Make sure to deploy your solution and provide a link in the PR description

etojeDenys (changes requested):
1. after pressing "Restart," the message about losing should disappear
image
2. it is better to use switch case here
image

vadiimvooo (approved):
Well done, just don't forget to add favicon

PR #1409, opened 2025-06-14
add: basic game functionality

gtailde:
- [DEMO LINK](https://gtailde.github.io/js_2048_game/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,6 @@
         </p>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
only add a new block if other blocks have been moved

If I move here down, a block will be added, but it shouldn't
image

etojeDenys (approved):
Great job!

PR #1412, opened 2025-06-16
add task solution

PastushenkoVik:
- [DEMO LINK](https://PastushenkoVik.github.io/js_2048_game/)


etojeDenys (changes requested):
1. you only need to add a new block if the others have moved. In this situation, if you move to the right, the block should not appear, but it does
image
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

2. The game can still be continued, but I see a message about losing.
image
image

etojeDenys (changes requested):
same problems. blocks spawn when the board doesn't change, and losing messages show up when there is a possible move

PastushenkoVik:
It looks like I forgot to  deploy
Sorry (

I fixed the message, but I can't catch any swamping blocks.

anastasiiavorobiova (approved):
Great work!

PR #1414, opened 2025-06-16
soluton

Radymyr:
[DEMO LINK](https://Radymyr.github.io/js_2048_game/)

Снимок экрана 2025-06-16 в 22 25 25


etojeDenys (changes requested):
1. the restart button should be styled differently. Center the word inside the button
image

2. when there is no legal move, a message about losing should be shown
image

3. zeros should not be displayed
image

volodymyr-soltys97 (changes requested):
Hi, the demo link looks broken, need to fix it 
image

Also, add a favicon to the page 
image

etojeDenys (approved):
Good job

PR #1444, opened 2025-07-08
add solution

Vl-Tsr:
[DEMO LINK](https://vl-tsr.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,7 +1,36 @@
...
+    game.moveRight(e.key);
+  }
+
+  if (e.key === 'ArrowUp') {
+    game.moveUp(e.key);
+  }
+
+  if (e.key === 'ArrowDown') {
+    game.moveDown(e.key);
+  }
```
etojeDenys, line 35: A switch case would be better here

etojeDenys (approved):
Good job

PR #1456, opened 2025-07-17
add task solution

isapq:
[DEMO LINK](https://isapq.github.io/js_2048_game/)


pedro-ruas (changes requested):
App isn't working properly.
When pressing start, nothing happens and console shows an error.

PR #1460, opened 2025-07-22
task.. done O_O

MykytaMusaiev:
[DEMO LINK](https://MykytaMusaiev.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -63,6 +63,10 @@ export default class Game {
       if (i + 1 < slidRow.length && slidRow[i] === slidRow[i + 1]) {
         const mergedValue = slidRow[i] * 2;
 
+        if (mergedValue === 2048) {
+          this.status = 'win';
+        }
+
         mergedRow.push(mergedValue);
         this.score += mergedValue;
 
@@ -133,7 +137,7 @@ export default class Game {
   }
 
   move(direction) {
-    if (this.status !== 'playing') {
+    if (this.status === 'game over') {
       return false;
     }
 
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -5,7 +5,7 @@
  * Now it has a basic structure, that is needed for testing.
  * Feel free to add more props and methods if needed.
  */
-class Game {
+export default class Game {
   /**
    * Creates a new game instance.
    *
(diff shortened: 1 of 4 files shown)


src/modules/Game.class.js
```diff
@@ -21,24 +21,176 @@ class Game {
...
+
+  _processRow(row) {
+    const slidRow = row.filter((cell) => cell !== 0);
+    const mergedRow = [];
+
+    for (let i = 0; i < slidRow.length; i++) {
+      if (i + 1 < slidRow.length && slidRow[i] === slidRow[i + 1]) {
+        const mergedValue = slidRow[i] * 2;
+
+        if (mergedValue === 2048) {
```
alexander-ignatow: 2048 is a magic number

src/modules/Game.class.js
```diff
@@ -21,24 +21,176 @@ class Game {
...
+        newBoard[j][i] = this.board[i][j];
+      }
+    }
+
+    return newBoard;
+  }
+
+  _checkForGameOver() {
+    for (let i = 0; i < 4; i++) {
+      for (let j = 0; j < 4; j++) {
```
alexander-ignatow: 4 is a magic number

even more - yes, classic 2048 is played on a 4x4 field, but  it makes sense to make this some sort of constant, not inline value (for instance - don't create board just explicitly as 4x4 matrix, rather make a change to create an NxN array, where N is provided externally)

src/modules/Game.class.js
```diff
@@ -21,24 +21,176 @@ class Game {
...
+          return;
+        }
+
+        if (i < 3 && this.board[i][j] === this.board[i + 1][j]) {
+          return;
+        }
+
+        if (j < 3 && this.board[i][j] === this.board[i][j + 1]) {
+          return;
+        }
```
alexander-ignatow: 3 looks like a magic number too

why the condition j<3 is so special?

src/modules/Game.class.js
```diff
@@ -21,24 +21,176 @@ class Game {
    * initial state.
    */
   constructor(initialState) {
-    // eslint-disable-next-line no-console
-    console.log(initialState);
+    this.score = 0;
+    this.status = 'idle'; // One of: 'idle', 'playing', 'win', 'lose'
```
alexander-ignatow: That is why you should not use JS for a modern real world tasks
status should be an enum (TS has support for enums and for union types)

Still, this is error prone to have a status like this - with simple literals

Step 1: add following code 

```
const Status = {
Idle: 'idle',
Playing: 'playing',
// ...
}
```

Step 2 

```suggestion
    this.status = Status.Idle;
```

See more on enums in JS: https://www.sohamkamani.com/javascript/enums/

src/modules/Game.class.js
```diff
@@ -21,24 +21,176 @@ class Game {
...
+        mergedRow.push(mergedValue);
+        this.score += mergedValue;
+
+        i++;
+      } else {
+        mergedRow.push(slidRow[i]);
+      }
+    }
+
+    while (mergedRow.length < 4) {
```
alexander-ignatow: the 4 looks as a magic number again

src/modules/Game.class.js
```diff
@@ -21,24 +21,176 @@ class Game {
...
+  }
+
+  move(direction) {
+    if (this.status === 'game over') {
+      return false;
+    }
+
+    const oldBoard = JSON.stringify(this.board);
+
+    switch (direction) {
```
alexander-ignatow, line 133: direction should be an enum too - see the Status

src/modules/Game.class.js
```diff
@@ -21,24 +21,176 @@ class Game {
...
+        this.board = this.board.map((row) => this._processRow(row));
+        break;
+      case 'Right':
+        this.board = this.board.map((row) => {
+          return this._processRow([...row].reverse()).reverse();
+        });
+        break;
+    }
+
+    const moved = JSON.stringify(this.board) !== oldBoard;
```
alexander-ignatow: That JSON trick looks like a hack to me

First - what exaclty do you gain with this optimisation?
Does it happen often, when the move results in no changes on the board?

If you still deside you need this, You should change this logic - maybe craft some sort of hash function for the state? Or just copy the internal state and compare it with the old?

alexander-ignatow: BTW - are you sure that you do save processor time creating and comparing JSON strings VS doing an "empty rerender"?

src/modules/Game.class.js
```diff
@@ -21,24 +21,176 @@ class Game {
...
+    return this.score;
+  }
 
   /**
    * @returns {number[][]}
    */
-  getState() {}
+  getState() {
+    return this.board;
+  }
```
alexander-ignatow: Change this to getters
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

src/modules/Game.class.js
```diff
@@ -50,19 +202,38 @@ class Game {
    * `win` - the game is won;
    * `lose` - the game is lost
    */
-  getStatus() {}
+  getStatus() {
+    return this.status;
+  }
```
alexander-ignatow: use a getter here too

src/modules/Game.class.js
```diff
@@ -50,19 +202,38 @@ class Game {
...
+    this.status = 'playing';
 
-  // Add your own methods here
-}
+    this.board = [
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+    ];
```
alexander-ignatow: I see this zeroes repeatedly.
Maybe you can add some function - like 

```
generateStartField = () => [
      [0, 0, 0, 0],
      [0, 0, 0, 0],
      [0, 0, 0, 0],
      [0, 0, 0, 0],
    ]
```

And use it anywhere you need this starting value

alexander-ignatow (changes requested):
Good, but lets make the code better

One more suggestion - it is out of the scope, but if I will just open the demo without reading the task, I cannot simply play the game
It would be nice to add some instructions bellow (like use arrow keys, etc)

src/modules/Game.class.js
alexander-ignatow: Step 3 - never compare the status with a literal directly - always use your object

`if (this.status === Status.Playing) ... `

MykytaMusaiev:
@alexander-ignatow HI!
Changed all  "magic numbers" to fixed values, updated UI, added new functionality, changed project code strucrute, fixed getters to be getters, changed empty field generation method, changed values to e-num type and hided inside of class, and added description before start, so as total move counter

ArturErmolenko (approved):
Looks great 🔥 Well done 👍

PR #1461, opened 2025-07-23
Develop

tavisua:
[DEMO LINK](https://tavisua.github.io/js_2048_game/)


etojeDenys (changes requested):
Make sure you link the style to the page then request a review again

anastasiiavorobiova (changes requested):
Great work!
The game adds many more numbers on the move than are needed

src/modules/Game.class.js
```diff
@@ -1,68 +1,374 @@
...
+      const column = this.state.map((row) => row[j]);
+      const filtered = column.filter((cell) => cell !== 0);
+      const merged = [];
+
+      for (let k = filtered.length - 1; k >= 0; k--) {
+        if (filtered[k] === filtered[k - 1]) {
+          const mergedValue = filtered[k] * 2;
+
+          merged.push(mergedValue);
+          this.score += mergedValue;
```
Anton-Kuchmasov, line 157: better to create a tiny helper functon (f.e. `addScore()`), which should sum prev score with mergedValue

Anton-Kuchmasov (approved):
Great work! Your solution is correct!

Please check out the comments — they might be useful for your future experience.

PR #1478, opened 2025-08-04
add task Solution

IvanR222:
- [DEMO LINK](https://ivanr222.github.io/js_2048_game/)


anastasiiavorobiova (changes requested):
The game adds much more value on restart than is needed.
Sometimes restart doesn't work
The game looks broken on the keypress

etojeDenys (changes requested):
the styles look broken. fix the path to them and then request the review again
image

Anton-Kuchmasov (approved):
Good job!

PR #1483, opened 2025-08-07
2048

Vhatever:
- [DEMO LINK](https://Vhatever.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,44 +1,58 @@
 'use strict';
 
-/**
- * This class represents the game.
- * Now it has a basic structure, that is needed for testing.
- * Feel free to add more props and methods if needed.
- */
 class Game {
-  /**
-   * Creates a new game instance.
-   *
-   * @param {number[][]} initialState
-   * The initial state of the board.
-   * @default
-   * [[0, 0, 0, 0],
-   *  [0, 0, 0, 0],
-   *  [0, 0, 0, 0],
-   *  [0, 0, 0, 0]]
-   *
-   * If passed, the board will be initialized with the provided
-   * initial state.
-   */
-  constructor(initialState) {
-    // eslint-disable-next-line no-console
-    console.log(initialState);
+  constructor(
+    initialState = [
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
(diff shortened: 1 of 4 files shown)


anastasiiavorobiova (changes requested):
Make sure to deploy your solution

Vhatever:
> Make sure to deploy your solution
тобто? щось не так задеплоїв?

src/modules/Game.class.js
```diff
@@ -1,44 +1,58 @@
...
-  moveRight() {}
-  moveUp() {}
-  moveDown() {}
+  moveLeft() {
+    if (this.isGameActive) {
+      this.move('left');
+    }
+  }
+  moveRight() {
+    if (this.isGameActive) {
```
TarasHoliuk: You do the same in each method. You can do it once in the previous function

src/modules/Game.class.js
```diff
@@ -50,19 +64,189 @@ class Game {
...
+
+        newRow = [...Array(zerosToAddEnd).fill(0), ...newRow];
+
+        return newRow;
+      });
+
+      return newTable;
+    };
+
+    const isGameOver = () => {
```
TarasHoliuk: Do not define this function in `move` method. Define it as a separate method. Check other similar cases

src/modules/Game.class.js
```diff
@@ -50,19 +64,189 @@ class Game {
...
+      }
+
+      this.placeNewCell();
+
+      this.isAbleMove = !isGameOver();
+    };
+
+    switch (direction) {
+      case 'up':
+        const moveUp = transpose(
```
TarasHoliuk: `moveUp` sounds like this is a function. It's a bit confusing. Rename it to reflect the value stored in this variable. Same below

TarasHoliuk (changes requested):
> тобто? щось не так задеплоїв?

Yep - try to open your demo link: https://vhatever.github.io/js_2048_game/
It doesn't work

Also added a few comments to improve your code

etojeDenys (changes requested):
after restarting the Start button has the same style as the restart button
image

Vhatever:
> after restarting the Start button has the same style as the restart button 

now it should work

maxim2310 (approved):
Well done!

PR #1508, opened 2025-08-23
add a solution

viktoriamyhailiak:
https://viktoriamyhailiak.github.io/js_2048_game/


vadiimvooo (changes requested):
Good job, few comments to fix. Please feel free to ask for help in fe_chat.
1. Game should start only when start button was pressed.
image
2. Remove lose message when game was restarted
image
3. Add favicon

vadiimvooo (changes requested):
Almost done, last problem remains. When no moves were done you should not create new cell. Example:
image
Here when I press bottom arrow I expect nothing will happen however it create new cell

etojeDenys (approved):
good job

PR #1510, opened 2025-08-24
2048 game

DmytroYakovchuk:
- [DEMO LINK](https://DmytroYakovchuk.github.io/js_2048_game/)


etojeDenys (changes requested):
1. fix game style before starting
image

2. center number
image

3. remove this text after starting the game
image

4. move up and down to spawn too many blocks

Anton-Kuchmasov (changes requested):
Good job!

However, there are a few issues that need to be fixed.

1. When I'm loading a page initially (first run, before game was started), layout seems to be broken:
image


2. Check out the comment from prev review - moving `down` or `up` adds more new cells than supposed. For example, here's game board after very first move - game added three cells at one time!
image

etojeDenys (approved):
good job

PR #1516, opened 2025-08-28
Solution for js_2048_game

tvk777:
- [DEMO LINK](https://tvk777.github.io/js_2048_game/)

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,29 @@
+name: Test
+
+on:
+  pull_request:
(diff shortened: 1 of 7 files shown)


etojeDenys (changes requested):
1. the user should not be able to make a move before the start button is pressed
image

2. A new block should only appear if another block was moved
before:
image

after: 
image

after the move to the right, a new block was spawned, but it shouldn't have been

vitalii-bondarenko-ua (approved):
Good job.

PR #1517, opened 2025-08-28
2047

DmytroPAvlichenko:
https://DmytroPAvlichenko.github.io/js_2048_game/

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,181 +1,227 @@
 'use strict';
 
-const tbody = document.querySelector('tbody');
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
image

Anton-Kuchmasov (changes requested):
Good job!

A couple issues that need to be fixed before approve:

1. When I'm starting a game, status should update immediately, not after 1st move (see screenshot)
image

2. Please remove `sdgg.js` file from repo - it's empty and has strange name

Anton-Kuchmasov (approved):

PR #1523, opened 2025-08-31
Develop

Voloshynd:
[DEMO LINK](https://Voloshynd.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,4 +1,4 @@
-/* eslint-disable prettier/prettier */
+/* eslint-disable function-paren-newline */
 'use strict';
 
 class Game {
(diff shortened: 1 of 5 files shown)


vitalii-fedusov:
image

Please fix the linter errors so that tests can be run on GitHub

src/modules/Game.class.js
```diff
@@ -20,25 +16,180 @@ class Game {
...
-    console.log(initialState);
+    this.initialState = Array.from({ length: Game.length }, () =>
+      Array(Game.length).fill(0));
+
+    this.sum = 0;
+    this.score = 0;
+    this.lastGeneratedCell = {};
+    this.startBtn = null;
+
+    this.field = document.querySelector('.game-field');
```
vitalii-fedusov: You should refer to DOM elements and work with them in the main.js file. Game.class should only contain logic.

src/modules/Game.class.js
```diff
@@ -50,19 +201,376 @@ class Game {
...
+      this.field.style.position = 'relative';
+      this.field.tBodies[0].append(this.createModal(str));
+    }
+  }
+
+  createModal(str) {
+    const overlay = document.createElement('div');
+    const overlayContent = document.createElement('div');
+
+    overlayContent.style.cssText = `
```
vitalii-fedusov: It is better to move the styling to a .css file.
In the JS code, when creating elements, simply add classes instead of style.cssText.
But again, it's better to work with DOM elements in the main.js file.

vitalii-fedusov (changes requested):
I really like your game!
The demo link looks great! Please do the refactor I asked for and fix the linter error so that the tests on GitHub will run.

src/modules/Game.class.js
Voloshynd: DOM elements moved to  main.js file

src/modules/Game.class.js
Voloshynd: styling moved to css file.

etojeDenys (approved):
good job

PR #1528, opened 2025-09-02
Solution

zaurmamedov:
- [DEMO LINK](https://zaurmamedov.github.io/js_2048_game/)


src/modules/Game.class.js
```diff
@@ -6,39 +6,224 @@
...
+      } else if (e.code === 'ArrowRight') {
+        this.moveRight();
+        this.setTwo();
+      } else if (e.code === 'ArrowUp') {
+        this.moveUp();
+        this.setTwo();
+      } else if (e.code === 'ArrowDown') {
+        this.moveDown();
+        this.setTwo();
+      }
```
etojeDenys: it would be better to use the switch case here

etojeDenys (changes requested):
the demo link does not work. fix it and then request the review
image

vadiimvooo (changes requested):
Good job. There are some fixes to do:
1. When restart button was clicked you should clear the board and game should not start instantly (only after clicking start)
2. Add favicon

etojeDenys (changes requested):
page looks broken now. fix it, and then request the review again
image

etojeDenys (changes requested):
almost done

make sure the user cannot play before pressing the start button
image

vitalii-bondarenko-ua (approved):
Looks good.

PR #1531, opened 2025-09-03
Develop

Kirky359:
- [DEMO LINK](https://Kirky359.github.io/js_2048_game/)

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,29 @@
+name: Test
+
+on:
+  pull_request:
+    branches: [ master ]
+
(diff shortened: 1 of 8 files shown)


etojeDenys (changes requested):
after pressing the start button you should replace it with restart. in your version, the restart button shows up after the first move so this can happen
image

Kirky359:
> after pressing the start button you should replace it with restart.

 - Fixed start button problem. 
 - [DEMO LINK](https://Kirky359.github.io/js_2048_game/)

etojeDenys (approved):
good job

PR #1535, opened 2025-09-08
2048 Game

Augustwise:
Since I already know how to use LocalStorage, I decided to implement it as a bonus feature for this project (instead of animation). Now, when the user reloads the browser page, their progress will always be saved and they won't lose it.

While writing the game, I also made visual notes for myself using a service called Excalidraw. There, I tried to describe the logic of row and column movements (which was not easy). If you want, you can view my visual notes at the following link: https://excalidraw.com/#json=NJd0LRcFzdqIxz1WGRJNN,gMIIaf8Djue4UxLL281HcQ

P.S. Unfortunately, while developing the game, I never managed to win in this game :(

[DEMO LINK](https://Augustwise.github.io/js_2048_game/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ You can change the HTML/CSS layout if you need it.
 ## Deploy and Pull Request
 
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_2048_game/)
+    - [DEMO LINK](https://Augustwise.github.io/js_2048_game/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
     - Run `npm run test:only -- -n` to run fast test ignoring linter;
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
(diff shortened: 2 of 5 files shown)


etojeDenys (changes requested):
it looks like you did a great job. to be sure, provide a demo link and then request a review again

Augustwise:
Oh, sorry

[DEMO LINK](https://Augustwise.github.io/js_2048_game/)

etojeDenys (changes requested):
to improve:

1. a new block should only appear if other blocks are moved.
in this situation, a new block should not appear after making a move up, but it did
image
image

2. after losing, there should be a restart button
image

3. after restarting, remove the message about losing
image

Augustwise:
I fixed the errors you mentioned.
"1. a new block should only appear if other blocks are moved" - I solved this problem in an interesting way: I compare the game field before pressing the movement button (using `JSON.stringify` to convert the field into a string) with the game field after pressing the button. If the field remains the same, I don’t add a new block to the game board.

Augustwise:
image


@etojeDenys I recommend deleting the gameState entry from LocalStorage before starting the game, so that there are no issues from previous attempts and the game runs correctly from a clean slate.

etojeDenys (approved):
great job

PR #1539, opened 2025-09-09
Solution

Pavlo-Dv:
[Demo](https://pavlo-dv.github.io/js_2048_game/)

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,29 @@
+name: Test
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


Pavlo-Dv:
image

etojeDenys (changes requested):
1. you should only show one button: the start button if the game has not started, and the restart button if it has
image

etojeDenys (changes requested):
after starting the game, the restart button appears, but the start button does not disappear. feel free to ask for help in the chat
image

etojeDenys (changes requested):
after losing, both buttons appear and the start button does not work. feel free to ask for help in the chat

image

etojeDenys (approved):
great job!

PR #1543, opened 2025-09-11
implemented base game requests

Metyck:
- [DEMO LINK](https://Metyck.github.io/js_2048_game/)


etojeDenys (changes requested):
1. a message about losing appears after the board is full, and the user's move does not work, but it is still possible to make a move
image

Anton-Kuchmasov (approved):
Well done!

PR #1548, opened 2025-09-13
Dev

fishxet:
[DEMO_LINK](https://fishxet.github.io/js_2048_game/)

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,29 @@
+name: Test
+
+on:
+  pull_request:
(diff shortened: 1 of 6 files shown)


etojeDenys (changes requested):
1. a new block should only appear if the other moves. after moving left, the new block swaps, but it should not
image
image

etojeDenys (changes requested):
problem is still not fixed
before move
image

after moving to the right
image

etojeDenys (approved):
good job

PR #1562, opened 2025-09-18
result

Vadym-Kit:
- [DEMO LINK](https://Vadym-Kit.github.io/js_2048_game/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,10 +1,11 @@
 'use strict';
 
-// // Uncomment the next lines to use your game instance in the browser
 const Game = require('../modules/Game.class');
 const game = new Game();
 
 const button = document.querySelector('.start');
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,7 +1,82 @@
...
+  if (evnt.key === 'ArrowDown') {
+    game.moveDown();
+  }
+
+  if (evnt.key === 'ArrowUp') {
+    game.moveUp();
+  }
+
+  render(game.getState());
+});
```
etojeDenys: it would be better to use switch case here

etojeDenys (changes requested):
1. after the game starts, remove this message
image

2. after restarting the game, remove the message about losing
image

etojeDenys (approved):
good job

PR #1572, opened 2025-09-28
add solution

dejucee:


etojeDenys (changes requested):

etojeDenys (changes requested):
problem is still not fixed. do not hesitate to ask for help in the chat
image

etojeDenys (changes requested):
1. after pressing the start button, the message should disappear and the entire game should shift down, but it does not
image

2. make the message about loss in the same style as the message at the start screen
image

vadiimvooo (approved):
Well done

PR #1577, opened 2025-09-30
solution

PODDN:
- [DEMO LINK](https://PODDN.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,7 +1,95 @@
...
+    game.moveRight();
+  }
+
+  if (evt.key === 'ArrowUp') {
+    game.moveUp();
+  }
+
+  if (evt.key === 'ArrowDown') {
+    game.moveDown();
+  }
```
etojeDenys, line 43: it would be better to use a switch case here

etojeDenys (approved):
good job

PR #1581, opened 2025-10-02
Game is ready

predator-bites:


src/scripts/main.js
```diff
@@ -1,7 +1,29 @@
...
+document.addEventListener('keydown', (e) => {
+  if (e.key === 'ArrowLeft') {
+    game.moveLeft(1);
+  } else if (e.key === 'ArrowRight') {
+    game.moveRight(1);
+  } else if (e.key === 'ArrowUp') {
+    game.moveUp(1);
+  } else if (e.key === 'ArrowDown') {
+    game.moveDown(1);
+  }
```
etojeDenys: it would be better to use a switch case here

etojeDenys (changes requested):
1. after trying to make a move before pressing the start buttons, we see a message about losing
image

2. after moving to the left in this position, the score increases
before move:
image

after:
image

etojeDenys (approved):
good job

predator-bites:
> good job

thanks, really appreciated to hear it

PR #1604, opened 2025-10-23
solution

tchuhai:
[DEMO LINK](https://tchuhai.github.io/js_2048_game/)


etojeDenys (changes requested):
great job

to improve:
after losing, there should be a restart button instead of a start button
image

vadiimvooo (approved):
Well done

PR #1610, opened 2025-10-31
created game 2048

VladStep186:
- [DEMO LINK](https://VladStep186.github.io/js_2048_game/)


src/modules/Game.class.js
```diff
@@ -22,23 +22,233 @@ class Game {
...
+    for (let i = 0; i < this.board.length; i++) {
+      const originalRow = [...this.board[i]];
+      let newRow = this.stackLine(originalRow);
+
+      newRow = this.slideAndMerge(newRow);
+      newRow = this.stackLine(newRow);
+
+      if (
+        !hasChanged &&
+        !newRow.every((cell, index) => cell === originalRow[index])
```
h-amster: move this code to a separate variable ablow and use it it here. Do it in all cases where condition in if takes more than 1 line

src/modules/Game.class.js
```diff
@@ -22,23 +22,233 @@ class Game {
...
+      this.board[i] = newRow;
+    }
+
+    if (hasChanged) {
+      this.addRandomTile();
+    }
+
+    this.checkWin();
+    this.checkLose();
+  }
```
h-amster, line 130: add an empty line between methods for better readability

h-amster (changes requested):
Good job! Almost done. Refactor you code according to my coments. Also error message from screenshot below should not be visible after restartin a game after lose

image

src/modules/Game.class.js
```diff
@@ -22,23 +22,239 @@ class Game {
...
+      );
+
+      if (!hasChanged && isColumnChanged) {
+        hasChanged = true;
+      }
+
+      for (let i = 0; i < this.board.length; i++) {
+        this.board[i][j] = newColumn[i];
+      }
+    }
```
brespect, line 175: Let`s make this like separate function:
> Reduce the number of passes over the column (combined the change check and writeback into one loop);
> Unify variable names (more readable);
> Add small optimizations and comments about options (e.g. early exit when the column is already empty).

src/scripts/main.js
```diff
@@ -1,7 +1,84 @@
...
+const scoreField = document.querySelector('.game-score');
+const messageLose = document.querySelector('.message-lose');
+const messageWin = document.querySelector('.message-win');
+const messageStart = document.querySelector('.message-start');
+const boardCells = document.querySelectorAll('.field-cell');
+
+function renderBoard() {
+  const state = game.getState();
+
+  for (let i = 0; i < 4; i++) {
```
brespect, line 16: avoid magic numbers, let`s add them to constants

src/scripts/main.js
```diff
@@ -1,7 +1,84 @@
...
+    for (let j = 0; j < 4; j++) {
+      const index = i * 4 + j;
```
brespect, line 18: check everywhere

brespect (changes requested):
Good job! Let`s improve logic:

etojeDenys (approved):
good job

PR #1636, opened 2025-11-25
Develop

st74eet:
- [DEMO LINK](https://st74eet.github.io/js_2048_game/)


etojeDenys (changes requested):
1. there is still a possible move, but the game is lost
image

2. new blocks should only appear if other blocks make a move
before move:
image

after: 
image

vadiimvooo (approved):
Well done

PR #1637, opened 2025-11-27
added solution

juliasemchishin:
https://juliasemchishin.github.io/js_2048_game/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -65,6 +65,9 @@
         </p>
       </div>
     </div>
-    <script src="scripts/main.js"></script>
+    <script
+      src="scripts/main.js"
+      type="module"
+    ></script>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
good job

1. after restarting, the message about losing should disappear
image

etojeDenys (changes requested):
demo does not work. please fix it and request the review again

juliasemchishin:
Hello.
I can not understand how I can fix it. Because I only added one line to the code and reloaded it. I tried to do it again, but it didn't help.

etojeDenys (changes requested):
1. demo now looks like this
image

if you have any questions or problems, it is better to ask for help in the chat because it is very uncomfortable to communicate here

brespect (approved):
Well done! As improvement you may add the animations

PR #1644, opened 2025-12-07
2048

horbenkoOleksandr:
[DEMO LINK](https://horbenkoOleksandr.github.io/js_2048_game/)


etojeDenys (changes requested):

Anton-Kuchmasov (changes requested):
Good job, but there are a few issues need to be fixed:

1. Would you like to making cells full-colored?
image
Just take a quick look [here](https://nabeyko.github.io/js_2048_game/)
you can use this solution as a reference & look up on which classnames you should add to cells:
image

2. The same suggestion is for `start/restart` button - it color should change depends on current game state

image

image

brespect (changes requested):
Almost done, but lets fix build errors first

horbenkoOleksandr:
що мається на увазі, не розумію?

Anton-Kuchmasov (approved):
Good job!

Before final review, try to fix all tests issues

PR #1655, opened 2025-12-16
Build 2048 game functionality

KaterynaHeraskevych:
In this project, I learned how to:

- Create a game class to store and manage the game state, score, and status

- Use a 2D array to represent the game board and update it after each move

- Implement movement logic for all directions (left, right, up, down)

- Handle tile merging rules and update the score correctly

- Generate new tiles randomly after valid moves

- Detect win and lose conditions based on the game state

- Connect game logic to the DOM by rendering the board and reacting to keyboard events

[DEMO LINK](https://KaterynaHeraskevych.github.io/js_2048_game/)


etojeDenys (changes requested):

Anton-Kuchmasov (changes requested):
Hey, something went wrong with game logic. For example, I can't commit a `down` move right now, but game position shows us that it should be allowed:
image

Try to fix your game' logic - if we'll take a look on previous commits, all tests were passed and game logic worked well

KaterynaHeraskevych:
> Hey, something went wrong with game logic. For example, I can't commit a `down` move right now, but game position shows us that it should be allowed: image
> 
> Try to fix your game' logic - if we'll take a look on previous commits, all tests were passed and game logic worked well

I've fixed the game logic. Could you check again please?

src/modules/Game.class.js
```diff
@@ -21,24 +16,245 @@ class Game {
...
+
+      for (let cellIndex = 1; cellIndex < this.size; cellIndex++) {
+        const cell = row[cellIndex];
+
+        if (cell > 0) {
+          let cellPosition = cellIndex;
+
+          // checking for available cell by
+          // traversal in opposite direction of current move
+          // and finding the most far empty cell to move
```
FaiHamid, line 50: Remove unnecessary comments

FaiHamid (approved):
Greate job!

PR #1684, opened 2026-01-04
added task solution

eg598:


eg598:
also, if you can, please give me tips on how I can animate this

etojeDenys (approved):
good job

you can add transitions to the blocks for animation, but to animate the movement of the blocks, you will need to rewrite some of the code

PR #1696, opened 2026-01-10
add task solution

maksymkolomiiets2-svg:
[DEMO LINK](https://maksymkolomiiets2-svg.github.io/js_2048_game/)


src/scripts/main.js
```diff
@@ -1,7 +1,87 @@
...
+
+  // Використовуємо camelCase назви методів
+  if (e.key === 'ArrowLeft') {
+    moved = game.moveLeft();
+  } else if (e.key === 'ArrowRight') {
+    moved = game.moveRight();
+  } else if (e.key === 'ArrowUp') {
+    moved = game.moveUp();
+  } else if (e.key === 'ArrowDown') {
+    moved = game.moveDown();
```
etojeDenys, line 69: it would be better to use a switch case here

etojeDenys (approved):
good job

PR #1715, opened 2026-01-21
2048 task

devTym:
- [DEMO LINK](https://devTym.github.io/js_2048_game/)


Anton-Kuchmasov (changes requested):
When game is started and then reseted, score should be 0:
image


Please fix it, re-deploy, add-commit-push your code and then re-request review again

etojeDenys (approved):
lgtm

PR #1723, opened 2026-01-28
Developer

Yar14k:


etojeDenys (changes requested):
good job

to fix:
1. restart button does not work

2. provide a demo link in the description

etojeDenys (changes requested):
after losing in this situation, the restart button does not work

Image

brespect (changes requested):
No DEMO Link provided.

Yar14k:
> No DEMO Link provided.

i have already

etojeDenys (changes requested):
problem is the same: the restart button doesn't work after loss. make sure you fix it before the next review
Image
button does not work in this situation

2pasha (approved):
lgtm 🔥

you need to add your demo link here ⬇️

Image

PR #1727, opened 2026-02-02
solved

vladahvala:
https://vladahvala.github.io/js_2048_game/


etojeDenys (changes requested):

etojeDenys (changes requested):
still does not work

Image

if you have any problems, feel free to ask for help in the chat room

etojeDenys (changes requested):
1. lose message appears after the first wrong move (in this case, moving to the left), but there are still possible moves

Image

2. after restarting the game, the score should reset as well
Image

vadiimvooo (approved):
Well done, just don't forget to add favicon

PR #1741, opened 2026-02-10
for check, one test fails - please say why

Natalia2408-forever:


brespect (changes requested):
Hello, seems like the test fails because the text does not appear in case of failure. Also, don`t forget to provide the DEMO LINK and pass all tests before PR request submit

etojeDenys (approved):
good job

PR #1749, opened 2026-02-16
2048

liena-grytsyna:
[DEMO LINK](https://liena-grytsyna.github.io/js_2048_game/)


brespect (changes requested):
Good progress, let's add the following:
- Add a background to the back of the board.
- Add animation of merging and moving squares.
- Add a favicon.
- Create a separate file with constants where you will put the magic values.

2pasha (approved):
lgtm 🔥

PR #1765, opened 2026-02-27
add task solution

naviailpach:


brespect (changes requested):
Good progress, let`s provide DEMO LINK to the PR's description bfore requesting the review

PR #1767, opened 2026-02-27
solution

UArtem87:
- DEMO LINK https://UArtem87.github.io/js_2048_game/

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ You can change the HTML/CSS layout if you need it.
 ## Deploy and Pull Request
 
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_2048_game/)
+    - [DEMO LINK](https://UArtem87.github.io/js_2048_game/)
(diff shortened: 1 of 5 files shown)


etojeDenys (changes requested):
good job, but there's one problem: the restart button should appear after pressing start instead of pressing buttons on a keyboard
Image

Image

UArtem87:
Change the Start button to Restart after the first move.

etojeDenys (changes requested):
problem has not been solved. make sure you deploy the changes

UArtem87:
Regarding the button state: the technical requirements (point [10]) state: Change the `Start` button to `Restart` after the first move..

My logic is as follows:

Clicking the "Start" button is the game initialization, not a "move" of the tiles itself.

If I change the button text immediately upon clicking "Start", I would technically violate the requirements, as no actual move has occurred yet.

Therefore, I implemented the button change specifically after the first keydown event that results in an actual board state change (a successful move).

If I am misinterpreting the term "first move," I would appreciate a clarification on how the team defines it. Thanks!"

etojeDenys (approved):

PR #1791, opened 2026-03-16
commiting just to see the feedback from AI since your damned tests do…

rustem-async:
… not run idk


Anton-Kuchmasov (changes requested):
Please deploy your solution, check if everything is working as expected, attach a DEMO link to the PR' description and then re-request review again

PR #1804, opened 2026-03-26
Develop

NataliaDavida:
- [DEMO LINK](https://NataliaDavida.github.io/js_2048_game/)


brespect (changes requested):
Good progress, but it seems like your deploy is broken:

brespect (changes requested):
Good progress, but it seems like your deploy is broken

PR #1808, opened 2026-03-30
task completed

VladChudin:
- [DEMO LINK](https://VladChudin.github.io/js_2048_game/)


etojeDenys (changes requested):
1. moves should not work before the game starts

Image

2. a new block should only appear after another move. in this situation, a new block appears after moving up, but it should not.
Image

Image

brespect (approved):
Good job!

PR #1839, opened 2026-04-21
Solution

ValentynaBahyrly:
[x] [DEMO LINK](https://ValentynaBahyrly.github.io/js_2048_game/)

[x] In this task, you need to implement the 2048 game like in [this reference](https://play2048.co/)

Okay, what do we have?
[x] 1) HTML and CSS are already written. You can use it, or implement your own design if you want.
[x] 2) Base `Game` class structure is already written too. Extend it with your own methods. Obligatory methods (used in tests):
  - constructor with `initialState` parameter (value is optional, defaults to the empty board)
  - `getState()`
  - `getScore()`
  - `getStatus()`
  - `moveLeft()`
  - `moveRight()`
  - `moveUp()`
  - `moveDown()`
  - `start()`
  - `restart()`

[x] 3) Reference.

That's it!

Okay, okay. Also, we have some rules:
[x] 1) The game field is 4 x 4
[x] 2) Each cell can be empty or contain one of the numbers: 2, 4, 8 ... 2^n
[x] 3) The player can move cells with keyboard arrows
[x] 4) All the numbers should be moved in the selected direction until all empty cells are filled in
[x]    - 2 equal cells should be merged into a doubled number
[x]    - The merged cell can’t be merged twice during one move
[x] 5) The move is possible if at least one cell is changed after the move
[x] 6) After move 2 or 4 appears in a random empty cell. 4 probability is 10%
[x] 7) When 2048 value is displayed in any cell, win message should be shown.
[x] 8) The `game over` message should be shown if there are no more available moves.
[x] 9) Hide start message when game starts.
[x] 10) Change the `Start` button to `Restart` after the first move.
[x] 11) `Restart` button should reset the game to the initial state.
[x] 12) Increase score with each move. The score should be increased by the sum of all merged cells.
[x] 13) The game consists of 2 main parts:
[x]   - game logic written in `src/modules/Game.class.js` module that exports `Game` class
[x]   - game UI written in `src/index.html` with `main.js` script that need to use `Game` class instance

[x]  Hints:
- You have class `field-cell--%cell_value%`, for styling cell in the game.
- Use `hidden` class for hiding elements on page.
- Use `start`, `restart` classes for the main button for different styles.
- Use `field-cell--%cell_value%` class like additional class, don't replace the main class.
- Use `keydown` event and `event.key` property to handle arrow buttons presses
    ```js
    document.addEventListener('keydown', event => console.log(event.key));
    ```
[x]  Adding animation to the game is optional. It is a bit tricky, but you can try it if you want. Probably, you will need to extend the Game class with additional methods and create a separate board storage with Tile entities to operate their corresponding DOM elements' positions.


etojeDenys (changes requested):
demo link does not work. please fix it and re-request the review

etojeDenys (changes requested):
1. restart button should appear after the start button is pressed. now, there can be a situation like this:

Image

2. game should start automatically after you press the restart button
Image

etojeDenys (changes requested):
there is still a problem with the restart button. it appears after one or two moves
Image

Image

brespect (approved):
Well done!

PR #1859, opened 2026-05-02
version 1

dianakomar:
https://dianakomar.github.io/js_2048_game/


etojeDenys (changes requested):
good job, to improve
1. restart button should appear after the start button is pressed. but in your solution, itappears after the first move

Image

2. also, messages disappear after making a move, even though I didn't press the start button
Image

etojeDenys (approved):
good job

PR #1897, opened 2026-05-31
add task solution

andre349:


andre349:
Hello!

I implemented the game logic and UI, but I ran into an issue with the project setup.

The tests fail before the game is actually tested. Parcel throws the following error:

"Browser scripts cannot have imports or exports"

The error points to:

const Game = require('../modules/Game.class');

As a result, the development server returns a 500 Internal Server Error and Cypress cannot open the page.

I tried different approaches (CommonJS and ES modules), but then either Parcel or the tests became incompatible.

Could you please clarify what module format is expected in this task and how Game.class.js should be connected to main.js in this project template?

Thank you.

vitalii-bondarenko-ua (changes requested):
Provide demo link in PR description.
Feel free to ask questions in chat.

andre349:
- [DEMO LINK](https://andre349.github.io/js_2048_game/)

andre349:
I figured out the issue: the browser could not work with `require` in `main.js`, so I changed it to ES module syntax and added `type="module"` in HTML.

I also fixed `getState()` to return a copied board instead of the original state reference.

The Parcel error was caused by stale cache, and `npm start` was failing because port 8080 was already used by an old Node process. After clearing the cache and stopping that process, everything worked. All tests are passing.

2pasha (changes requested):
your app doesn't work 

Image

you need to fix paths 

Image

if you have some troubles - feel free to ask your question in fe_chat 💬

andre349:
- [DEMO LINK](https://andre349.github.io/js_2048_game/)

Anton-Kuchmasov (approved):
Well done!

PR #1907, opened 2026-06-13
add task solution  js_2048_game

zarubina20:


Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. After pressing `Restart` button the game should be restarted. `Start` button is better to appear only when page is loaded or reloaded.
Image

brespect (approved):
Good job!

PR #1916, opened 2026-06-24
Implement 2048 game logic

Annybianzinny:
## Demo

https://annybianzinny.github.io/js_2048_game/

## Test Report

https://annybianzinny.github.io/js_2048_game/report/

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -1,68 +1,201 @@
 'use strict';
 
-/**
- * This class represents the game.
- * Now it has a basic structure, that is needed for testing.
- * Feel free to add more props and methods if needed.
- */
 class Game {
-  /**
-   * Creates a new game instance.
-   *
-   * @param {number[][]} initialState
-   * The initial state of the board.
-   * @default
-   * [[0, 0, 0, 0],
-   *  [0, 0, 0, 0],
-   *  [0, 0, 0, 0],
-   *  [0, 0, 0, 0]]
-   *
-   * If passed, the board will be initialized with the provided
-   * initial state.
-   */
-  constructor(initialState) {
-    // eslint-disable-next-line no-console
-    console.log(initialState);
+  constructor(
+    initialState = [
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+    ],
+  ) {
+    this.initialState = initialState.map((row) => [...row]);
+    this.board = initialState.map((row) => [...row]);
+
+    this.score = 0;
+    this.status = 'idle';
   }
 
-  moveLeft() {}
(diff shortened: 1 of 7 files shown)


raulriato (changes requested):
Your task is not done. It seems you didn't `push` any code

raulriato (changes requested):
You didn't follow the [Layout Task Guideline](https://mate-academy.github.io/layout_task-guideline/#how-to-solve-the-layout-tasks-on-github).

Your `DEMO LINK` is not in the `pull request description` and it doesn't even workl. Also, you didn't update the `readme.md`.

Please make sure your `DEMO LINK` works before re-requesting review.

src/scripts/main.js
```diff
@@ -1,7 +1,98 @@
 'use strict';
 
-// Uncomment the next lines to use your game instance in the browser
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+/* global Game */
 
-// Write your code here
+const game = new Game();
```
raulriato: This is wrong. You should have just uncommented the lines, but now you are not importing the Game class into this file

src/modules/Game.class.js
```diff
@@ -1,68 +1,201 @@
...
+  moveDown() {
+    this.makeMove(
+      (board) => this.reverse(this.transpose(board)),
+      (board) => this.transpose(this.reverse(board)),
+    );
+  }
 }
 
-module.exports = Game;
+window.Game = Game;
```
raulriato: This is not the way to export the class. Return it to `module.exports`

src/scripts/main.js
```diff
@@ -1,7 +1,98 @@
...
+  if (game.getStatus() === 'lose') {
+    loseMessage.classList.remove('hidden');
+  }
+}
+
+button.addEventListener('click', () => {
+  if (game.getStatus() === 'idle') {
+    game.start();
+
+    button.textContent = 'Restart';
```
raulriato: This change should occur after the player makes theis first move, not when they click `start`

src/modules/Game.class.js
raulriato, line 1: This is not a problem, just a tip.

To have a better code, you could have your class separated by public methods, the ones that will be called outside the class (`start`, `moveUp`, `moveDown`, `restart`, etc.) and helper methods, that are only supposed to help the public methods (`makeMove`, `addRandomCell`, etc.)

This is not a necessary change, but since I'm asking for some important changes, it would be nice to make this as well.

raulriato (changes requested):
Your `start` button does not work because of a tiny error. If you fix it, I believe you'll be good to go.

Remember to rerun `npm run deploy` after pushing your changes to make sure the `DEMO LINK` keps up with the code.

DanielCamposDevX (approved):

PR #1925, opened 2026-07-02
add task solution

ybriukhovetska:
- [DEMO LINK](https://ybriukhovetska/js_2048_game/)


vadiimvooo (changes requested):
Good job, there are some fixes to do:
1. You should remove start message when game was started
Image

2. After restart lose message should be removed
Image

etojeDenys (changes requested):
almost done

1: after losing, the restart button should appear instead of the start button
Image

etojeDenys (approved):
great job

PR #1944, opened 2026-07-23
solution

mariana-zh:
- [DEMO LINK](https://mariana-zh.github.io/js_2048_game/)


etojeDenys (changes requested):
good job, just one fix:
- disable the moves before pressing the start button because this can happen

Image

etojeDenys (approved):
lgtm

PR #1957, opened 2026-08-06
solution: game 2048

AlexanderOkhotnichenko:
- [DEMO LINK](https://alexanderokhotnichenko.github.io/js_2048_game/)


src/modules/Game.class.js
```diff
@@ -21,24 +21,264 @@ class Game {
...
+        continue;
+      }
+
+      if (
+        index < filterLine.length - 1 &&
+        filterLine[index] === filterLine[index + 1]
+      ) {
+        result.push(filterLine[index] * 2);
+        score += filterLine[index] * 2;
+        skipCell = true;
```
Zoolgrand: Double skip — index++ already skips the merged tile, so this breaks [2,2,2,2] → [4,2,2]. Remove skipCell entirely, keep only index++.

src/modules/Game.class.js
```diff
@@ -21,24 +21,264 @@ class Game {
...
+    }
+
+    return moved;
+  }
+
+  moveRight() {
+    let moved = false;
+
+    for (let row = 0; row < 4; row++) {
+      const reversed = this.board[row].reverse();
```
Zoolgrand: .reverse() mutates the board, and result.line gets reversed twice → right move ends up left-aligned.

src/modules/Game.class.js
```diff
@@ -21,24 +21,264 @@ class Game {
    * initial state.
    */
   constructor(initialState) {
-    // eslint-disable-next-line no-console
-    console.log(initialState);
+    this.initialState = initialState;
+
+    this.board =
+      initialState ||
```
Zoolgrand: Shared reference — mutates the caller's array. Use clone: initialState.map(row => [...row]). Same in restart().

Zoolgrand (changes requested):
game looks good overall but I've found a couple of bugs. Could you take a look?

brespect (changes requested):
When I click the right arrow, the merging of the squares looks broken:
Image

etojeDenys (approved):
good job

PR #1972, opened 2026-08-25
add task solution

Artroy666:
- [DEMO LINK](https://Artroy666.github.io/js_2048_game/)
Tried doing animations but ended up being too hard for me but its optional so should be ok, everything else should work perfectly.
Was not the hardest task, got kinda confused by comments and base things in files in the start (on me probably), but everything should be good in the end


brespect (approved):
Well done!

PR #1985, opened 2026-09-03
2048 game solution

polina3107:
- [DEMO LINK](https://polina3107.github.io/js_2048_game/)


etojeDenys (changes requested):
good job, to improve:
1. currently, new blocks appear when any button is pressed, but this should not happen

2. make sure that new blocks only appear when another block moves

vadiimvooo (approved):
Well done

PR #1986, opened 2026-09-04
game

notbelievable:
- [DEMO LINK](https://notbelievable.github.io/js_2048_game/)

### DIFF
--- a/src/modules/Game.class.js
+++ b/src/modules/Game.class.js
@@ -23,22 +23,130 @@ class Game {
   constructor(initialState) {
     // eslint-disable-next-line no-console
     console.log(initialState);
+    this.initialState = initialState;
+    this.score = 0;
+    this.status = 'idle';
+
+    this.board = initialState || [
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+      [0, 0, 0, 0],
+    ];
   }
 
-  moveLeft() {}
-  moveRight() {}
-  moveUp() {}
-  moveDown() {}
+  shiftLeft() {
+    const noZero = this.board.map((str) => {
+      return str.filter((num) => num !== 0);
+    });
+
+    for (let i = 0; i < 4; i++) {
+      let str = noZero[i];
+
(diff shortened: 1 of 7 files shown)


brespect (changes requested):
Sorry, but your DEMO LINK is not working:
Image

src/modules/Game.class.js
```diff
@@ -23,22 +23,130 @@ class Game {
...
+      this.addRandomTile();
+    }
+
+    const flatBoard = this.board.flat();
+
+    if (flatBoard.includes(2048)) {
+      this.status = 'win';
+    } else if (!flatBoard.includes(0)) {
+      this.status = 'lose';
+    }
```
Zoolgrand, line 79: This check only in this branch now, so we can wit or lose only when we move left.

src/modules/Game.class.js
```diff
@@ -50,19 +158,55 @@ class Game {
...
+
+    if (nums.length === 0) {
+      return;
+    }
+
+    const randomInd = Math.floor(Math.random() * nums.length);
+    const randomCell = nums[randomInd];
+
+    this.board[randomCell.row][randomCell.cell] = 2;
```
Zoolgrand: We need to also include 10% chance to start with "4"

src/scripts/main.js
```diff
@@ -1,7 +1,92 @@
...
-// const Game = require('../modules/Game.class');
-// const game = new Game();
+const Game = require('../modules/Game.class');
+const game = new Game();
 
 // Write your code here
+const mesStart = document.querySelector('.message-start');
+const butStart = document.querySelector('.start');
+
+butStart.addEventListener('click', (e) => {
```
Zoolgrand: (e) is unused we can remove this parameter

Zoolgrand (changes requested):
Good progress, you need to fix 2 issues related to game logic and 1 minor issue

roman-mirzoian (approved):
Well done 👍

Zoolgrand (approved):

